PR #15, opened 2018-03-07
Split documentation into multiple files

greg5green:
## Why?
Having all of the generated documentation in one file was a bit unwieldy. By splitting the docs into multiple files, it's a lot easier to parse.

## What changed?
- Introduced custom build script that builds out:
  - An index of classes and typedefs
  - An individual file for each class
  - One file with all typedefs
- Created custom helpers and partials to have links between pages go to the correct places


support/docs/partials/sig-link-html.hbs
```diff
@@ -0,0 +1,14 @@
+{{#if name}}{{#sig no-gfm=true ~}}
+{{{@depOpen}~}}
+<a href="{{{readmeLink}}}">
```
zachary-kuhn: I looked through their template code and this looks like the best way. Great work figuring this out, @greg5green!

PR #18, opened 2018-03-15
Add Facilities#create, Facilities#delete, Facilities#getAllByOrganizationId, Facilities#update

greg5green:
## Why?
There is additional functionality in the facilities manager API that we need to be able to do through the SDK.

## What changed?
- Added Facilities#create that creates a new facility
- Added Facilities#delete that deletes an existing facility
- Added Facilities#getAllByOrganizationId that gets all facilities that belong to a particular organization
- Added Facilities#update that updates an existing facility
- Started normalizing data coming from the facilities manager API
  - There are some inconsistencies with how things are named in facilities responses

### DIFF
--- a/docs/Facilities.md
+++ b/docs/Facilities.md
@@ -10,6 +10,7 @@ of, information about different facilities
     * [new Facilities(sdk, request)](#new_Facilities_new)
     * [.get(facilityId)](#Facilities+get) ⇒ <code>Promise</code>
     * [.getAll()](#Facilities+getAll) ⇒ <code>Promise</code>
+    * [.getAllByOrganizationId(organizationId)](#Facilities+getAllByOrganizationId) ⇒ <code>Promise</code>
 
 <a name="new_Facilities_new"></a>
 
@@ -56,6 +57,28 @@ Method: GET
 **Example**  
 ```js
 contxtSdk.facilities.getAll()
+  .then((facilities) => console.log(facilities));
+  .catch((err) => console.log(err));
+```
+<a name="Facilities+getAllByOrganizationId"></a>
+
+### contxtSdk.facilities.getAllByOrganizationId(organizationId) ⇒ <code>Promise</code>
+Gets a list of all facilities that belong to a particular organization
+
+API Endpoint: '/organizations/:organizationId/facilities'
+Method: GET
+
+**Kind**: instance method of [<code>Facilities</code>](#Facilities)  
+**Fulfill**: <code>Facility[]</code> Information about all facilities  
+**Reject**: <code>Error</code>  
+
+| Param | Type |
+| --- | --- |
+| organizationId | <code>number</code> \| <code>string</code> | 
+
+**Example**  
+```js
+contxtSdk.facilities.getAllByOrganizationId(25)
   .then((facilities) => console.log(facilities));
   .catch((err) => console.log(err));
 ```
--- a/docs/Facilities.md
+++ b/docs/Facilities.md
@@ -13,6 +13,7 @@ of, information about different facilities
     * [.get(facilityId)](#Facilities+get) ⇒ <code>Promise</code>
     * [.getAll()](#Facilities+getAll) ⇒ <code>Promise</code>
     * [.getAllByOrganizationId(organizationId)](#Facilities+getAllByOrganizationId) ⇒ <code>Promise</code>
+    * [.update(facilityId, update)](#Facilities+update) ⇒ <code>Promise</code>
 
 <a name="new_Facilities_new"></a>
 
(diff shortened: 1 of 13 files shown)


docs/Facilities.md
```diff
@@ -59,3 +120,50 @@ contxtSdk.facilities.getAll()
...
+Method: PUT
+
+**Kind**: instance method of [<code>Facilities</code>](#Facilities)  
+**Fulfill**: <code>undefined</code>  
+**Reject**: <code>Error</code>  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| facilityId | <code>number</code> | The id of the facility to update |
+| update | <code>Object</code> | An object containing the updated data for the facility |
```
rob5standingby: Do you want to list out all of the possible things that could be updated?

rob5standingby: This may be listed out elsewhere in the documentation; I'm not quite sure

src/facilities.js
```diff
@@ -41,13 +44,84 @@ class Facilities {
...
+   *     name: 'Sherlock Homes Museum',
+   *     organizationId: 25
+   *   })
+   *   .then((facilities) => console.log(facilities));
+   *   .catch((err) => console.log(err));
+   */
+  create(facility = {}) {
+    const requiredFields = ['organizationId', 'name', 'timezone'];
+
+    for (let i = 0; requiredFields.length > i; i++) {
```
rob5standingby: Is there any reason you're iterating over the array in that manner instead of `['organizationId', 'name', 'timezone'].forEach()`? Just curious if there was a compatibility concern. I do think the `for()` looks cleaner, but just curious!

src/sessionTypes/auth0WebAuth.js
```diff
@@ -119,7 +119,13 @@ class Auth0WebAuth {
               reject(err);
             }
 
-            resolve(profile);
+            const formattedProfile = {
+              ...profile,
```
rob5standingby: Is there a `created_at` field that should be handled similarly?

src/sessionTypes/auth0WebAuth.spec.js
```diff
@@ -196,7 +199,7 @@ describe('sessionTypes/Auth0WebAuth', function() {
 
       it("returns a fulfilled promise with the users's profile", function() {
         return expect(promise).to.be.fulfilled
-          .and.to.eventually.equal(expectedProfile);
+          .and.to.eventually.deep.equal(expectedProfile);
```
rob5standingby: I was not aware of `to.be.fulfilled.and.to.eventually.equal` 💯

support/fixtures/factories/facility.js
```diff
@@ -5,36 +5,64 @@ const faker = require('faker');
...
     return times(faker.random.number({ min: 0, max: 5 }), () => {
-      return factory.build('facilityTag', { facility_id: id });
+      return factory.build('facilityTag', { facilityId: id }, { fromServer });
     });
+  })
+  .after((facility, options) => {
+    // If building a facility object that comes from the server, transform it to have camel case and
+    // capital letters in the right spots
+    if (options.fromServer) {
+      facility.created_at = facility.createdAt;
```
rob5standingby: Doesn't `updated_at` also get returned?

support/fixtures/factories/facility.js
```diff
@@ -5,36 +5,64 @@ const faker = require('faker');
...
 
 factory.define('facility')
+  .option('fromServer', false)
   .sequence('id')
   .attrs({
     address1: () => faker.address.streetAddress(),
     address2: () => faker.address.secondaryAddress(),
     city: () => faker.address.city(),
-    Info: () => {
+    createdAt: () => faker.date.past().toISOString(),
```
rob5standingby: Do you care about having `updatedAt` here?

docs/Facilities.md
greg5green: Oh yeah, good idea. I did it for create, but not update

src/facilities.js
greg5green: Using a for loop in this case because when you return, it returns `create` instead of the anonymous function that you'd be using in forEach. You could make it work with forEach or map, but it would be a bit more code

src/sessionTypes/auth0WebAuth.js
greg5green: Not from auth0

src/sessionTypes/auth0WebAuth.spec.js
greg5green: It is part of `chai-as-promised`

support/fixtures/factories/facility.js
greg5green: I'm not saying that it cannot be part of this data, but there isn't a single example of that has it on staging -- is there on production?

rob5standingby: Damn. Really!! One sec...

jongamble:
I really like how clean this is. Great job! :shipit:

support/fixtures/factories/facility.js
rob5standingby: Yeah, double-checked too. No `updated_at` gets returned; even though the data is stored in Postgres. Carry on!

greg5green: 👍 We can definitely add it in the future if we need it

rob5standingby:
:shipit: !!!

PR #19, opened 2018-03-28
Add Facilities#updateInfo

greg5green:
## Why?
Some Facility attributes like `square_feet` are stored as facility info instead of directly on the facility model. We want to update those attributes.

## What changed?
- Added Facilities#updateInfo method that POSTs to `/facilities/:facilityId/info`

### DIFF
--- a/src/facilities.js
+++ b/src/facilities.js
@@ -244,6 +244,42 @@ class Facilities {
   }
 
   /**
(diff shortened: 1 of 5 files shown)


greg5green:
@rob5standingby Yeah. Always getting a 204 at the moment. Seems to create an initial `square_feet` value just fine in `facility_info`, but is having trouble updating it. Digging into it here in a little bit!

PR #20, opened 2018-03-28
Rename Facilities#updateInfo to Facilities#createOrUpdateInfo

greg5green:
## Why?
1. The method name implied it would be doing a PATCH instead of a POST, and clarity is good.
2. An additional query param is required to be sent to the API for this to behave correctly.

## What changed?
- Renamed the method
- Added the `should_update` query param to the request

### DIFF
--- a/src/facilities.js
+++ b/src/facilities.js
@@ -94,6 +94,48 @@ class Facilities {
   }
 
   /**
+   * Creates or updates a facility's info (NOTE: This refers to the facility_info model)
+   *
+   * API Endpoint: '/facilities/:facilityId/info?should_update=true'
+   * Method: POST
+   *
+   * @param {number} facilityId The id of the facility to update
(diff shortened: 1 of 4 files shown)


src/facilities.js
```diff
@@ -93,6 +93,48 @@ class Facilities {
...
+  createOrUpdateInfo(facilityId, update) {
+    if (!facilityId) {
+      return Promise.reject(new Error("A facility id is required to update a facility's info."));
+    }
+
+    if (!update) {
+      return Promise.reject(new Error("An update is required to update a facility's info."));
+    }
+
+    if (!isPlainObject(update)) {
```
zachary-kuhn: Is an empty `update` of `{}` also valid?

greg5green: It won't error, but it still will make a request. So, yes.

PR #21, opened 2018-04-06
Facilities#createGrouping

greg5green:
## Why?
We need the ability to create new facility groupings.

## What changed?
- Added `Facilities@createGrouping`
- Moved the facility transform utils/helpers to a utils directory
  - Cleans up the Facilities class quite a bit
  - Future improvement would be figuring out how to DRY these up a bit. Potentially could pull in a package that does camelCase -> snake_case and vice versa. May need to build an adapter for it or do a custom implementation since some keys we will not want to transform, or will require custom transforms

### DIFF
--- a/src/utils/facilities/formatFacilityToServer.spec.js
+++ b/src/utils/facilities/formatFacilityToServer.spec.js
@@ -1,0 +1,37 @@
+import omit from 'lodash.omit';
+import formatFacilityToServer from './formatFacilityToServer';
+
+describe('utils/facilities/formatFacilityToServer', function() {
+  let expectedFacility;
+  let facility;
+  let formattedFacility;
+
+  beforeEach(function() {
+    facility = fixture.build('facility');
+    expectedFacility = omit(
+      {
+        ...facility,
+        geometry_id: facility.geometryId,
+        Info: facility.info,
+        organization_id: facility.organizationId,
+        weather_location_id: facility.weatherLocationId
+      },
+      [
+        'createdAt',
+        'geometryId',
+        'id',
+        'info',
+        'organization',
+        'organizationId',
+        'tags',
+        'weatherLocationId'
+      ]
+    );
+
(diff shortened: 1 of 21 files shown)


src/utils/facilities/formatFacilityToServer.spec.js
```diff
@@ -0,0 +1,37 @@
...
+import formatFacilityToServer from './formatFacilityToServer';
+
+describe('utils/facilities/formatFacilityToServer', function() {
+  let expectedFacility;
+  let facility;
+  let formattedFacility;
+
+  beforeEach(function() {
+    facility = fixture.build('facility');
+    expectedFacility = omit(
```
zachary-kuhn: It feels like this approach is frail and can be broken fairly easily by unrelated changes, but I have no suggestions on how to improve it.

support/fixtures/factories/facilityGrouping.js
```diff
@@ -0,0 +1,41 @@
...
+    name: () => faker.commerce.productName(),
+    organizationId: () => factory.build('organization').id,
+    ownerId: () => faker.internet.userName(),
+    parentGroupingId: () => faker.random.uuid(),
+    updatedAt: () => faker.date.recent().toISOString()
+  })
+  .after((facilityGrouping, options) => {
+    // If building a facility grouping object that comes from the server, transform it to have camel
+    // case and capital letters in the right spots
+    if (options.formServer) {
```
zachary-kuhn: Is `formServer` a typo?

greg5green: Ugh, yes.

src/utils/facilities/formatFacilityToServer.spec.js
greg5green: I don't really like it much either but haven't come up with a great way to handle it. Tbh, even coming up with doing it this way was kind of a challenge. If you think of a good idea, lmk

PR #22, opened 2018-04-10
Add Facility to Facility Grouping

greg5green:
## Why?
Facility Groupings aren't very useful without facilities in them.

## What changed?
- Added method to add a facility to a facility grouping
- Created separate FacilityGroupings class that is accessed like `contxtService.facilities.groupings.methodName`
  - Facilities class was getting large and a little unweildly (especially with all that JSDoc)
  - FacilityGroupings are still part of the Facilities service, so did not necessarily want to move it out entirely to its own namespace

### DIFF
--- a/src/facilities/groupings.js
+++ b/src/facilities/groupings.js
@@ -66,9 +66,9 @@ class FacilityGroupings {
     let errorMsg;
 
     if (!facilityGroupingId) {
-      errorMsg = 'A facilityGroupingId is required to create a between a facility grouping and a facility.';
+      errorMsg = 'A facilityGroupingId is required to create a relationship between a facility grouping and a facility.';
     } else if (!facilityId) {
-      errorMsg = 'A facilityId is required to create a between a facility grouping and a facility.';
+      errorMsg = 'A facilityId is required to create a relationship between a facility grouping and a facility.';
     }
 
     if (errorMsg) {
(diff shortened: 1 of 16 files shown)


src/facilities/groupings.js
```diff
@@ -0,0 +1,131 @@
...
+   * @example
+   * contxtSdk.facilities.groupings.addFacility('b3dbaae3-25dd-475b-80dc-66296630a8d0', 4)
+   *   .then((grouping) => console.log(grouping));
+   *   .catch((err) => console.log(err));
+   */
+  addFacility(facilityGroupingId, facilityId) {
+    let errorMsg;
+
+    if (!facilityGroupingId) {
+      errorMsg = 'A facilityGroupingId is required to create a between a facility grouping and a facility.';
```
zachary-kuhn: The wording is off a little here.

zachary-kuhn: And the one below!

src/facilities/groupings.js
```diff
@@ -0,0 +1,131 @@
...
+   * @fulfill {FacilityGrouping} Information about the new facility grouping
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.facilities.groupings
+   *   .create({
+   *     description: 'US States of CT, MA, ME, NH, RI, VT',
+   *     isPrivate: false,
+   *     name: 'New England, USA',
+   *     organization_id: '61f5fe1d-d202-4ae7-af76-8f37f5bbeec5'
```
zachary-kuhn: The documentation above uses camelcase, but this uses snake case. 🐍

greg5green: Good catch!

src/facilities/groupings.js
greg5green: 😩

PR #25, opened 2018-04-16
Add a Wercker config

greg5green:
## Why?
Having some form of automated testing/building/linting will be helpful to our CI process (and might eventually get us to CD on this project).

## What changed?
- Added a config for Wercker


wercker.yml
```diff
@@ -0,0 +1,12 @@
+box: node:8
+
+build:
+  steps:
+    - npm-install:
+      use-cache: false
```
zachary-kuhn: Never use cache? Won't this make the build slower than it needs to be?

PR #28, opened 2018-05-01
Task/facility groupings update

greg5green:
## Why?
We're working our way through CRUD on FacilityGroupings.

~~C R U D~~

## What changed?
- Added FacilityGrouping#update method


greg5green:
I don't know if that's good, Zach.

PR #30, opened 2018-05-17
Add Prettier

greg5green:
## Why?
We've been using Prettier on our other projects and it's pretty great.

## What changed?
- Added Prettier
  - Runs on a precommit hook 💯 
  - ESLint rules are updated to match/ignore rules that are related to what Prettier does
- Ran Prettier on existing project files so future commits aren't crazy


greg5green:
@zachary-kuhn Thankfully, it's almost all machine generated.

PR #32, opened 2018-06-11
FM-91: Add cost centers for the facility formatter

leeallen337:
## Why?

Need to see the cost centers linked to a facility

https://ndustrialio.atlassian.net/browse/FM-91

## What changed?
- `Facilities#getAllByOrganizationId` to accept parameters to include cost centers information
- `Facilities#get` to include cost centers information for that specific facility

### DIFF
--- a/src/facilities/index.js
+++ b/src/facilities/index.js
@@ -267,12 +267,6 @@ class Facilities {
       );
     }
 
-    if (!isPlainObject(options)) {
-      return Promise.reject(
-        new Error('The options must be a well-formed object.')
-      );
-    }
-
     const params = formatFacilityOptionsToServer(options);
 
     return this._request
--- a/src/facilities/index.js
+++ b/src/facilities/index.js
@@ -267,6 +267,12 @@ class Facilities {
       );
     }
 
+    if (!isPlainObject(options)) {
+      return Promise.reject(
+        new Error('The options must be a well-formed object.')
+      );
+    }
+
     const params = formatFacilityOptionsToServer(options);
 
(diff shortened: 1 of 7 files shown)


src/facilities/index.js
```diff
@@ -267,6 +267,12 @@ class Facilities {
       );
     }
 
+    if (!isPlainObject(options)) {
```
zachary-kuhn: What if passed undefined? It's usually OK to not pass an options object to a call for most libraries.

src/utils/facilities/formatFacilityOptionsToServer.spec.js
```diff
@@ -29,6 +31,32 @@ describe('utils/facilities/formatFacilityOptionsToServer', function() {
     expect(options).to.deep.equal(expectedOptions);
   });
 
+  it('provides an object that includes the `include_cost_centers` that the API service can understand', function() {
```
zachary-kuhn: Maybe missing a word? This didn't read quite fluidly to me.

leeallen337: 👍 

Added a word that hopefully helps

src/facilities/index.js
greg5green: I agree with this comment.

zachary-kuhn: Default parameter is good! Needs less error checking that way.

leeallen337:
@zachary-kuhn & @greg5green if you don't have anymore comments could you approve this?

PR #34, opened 2018-06-13
Feature/cost center crud

jongamble:
## Why?
We need the ability to interact with and manipulate the Cost Center data from the API.

## What changed?
`CostCenters#addFacility` to add a facility to a cost center
`CostCenters#create` for creating a new cost center
`CostCenters#getAll` for getting a list of all cost centers
`CostCenters#getAllByOrganizationId` for getting all cost centers for a specific organization
`CostCenters#remove` to remove an existing cost center
`CostCenters#removeFacility` to remove a facility from a cost center
`CostCenters#update` to update an existing cost center

### DIFF
--- a/src/facilities/costCenters.js
+++ b/src/facilities/costCenters.js
@@ -6,12 +6,6 @@ import {
 } from '../utils/facilities';
 
 /**
- * Module that provides access to cost centers, and helps manage
- * the relationship between those cost centers and facilities
- *
- * @typicalname contxtSdk.facilities.costCenters
- */
-/**
  * @typedef {Object} CostCenter
  * @param {string} createdAt ISO 8601 Extended Format date/time string
  * @param {string} [description]
@@ -20,6 +14,7 @@ import {
  * @param {string} organizationId UUID
  * @param {string} updatedAt ISO 8601 Extended Format date/time string
  */
+
 /**
  * @typedef {Object} CostCenterFacility
  * @param {string} costCenterId UUID
@@ -27,6 +22,13 @@ import {
  * @param {number} facilityId
  * @param {string} id UUID
  * @param {string} updatedAt ISO 8601 Extended Format date/time string
+ */
+
+/**
+ * Module that provides access to cost centers, and helps manage
+ * the relationship between those cost centers and facilities
+ *
+ * @typicalname contxtSdk.facilities.costCenters
  */
 class CostCenters {
   /**
@@ -262,8 +264,8 @@ class CostCenters {
    * @example
    * contxtSdk.facilities.costCenters
    *   .update({
-   *     description: 'US States of CT, MA, ME, NH, RI, VT',
-   *     name: 'New England, USA',
+   *     description: 'Refrigeration compressors throughout the facility',
+   *     name: 'Compressors',
    *   })
    *   .then((costCenter) => console.log(costCenter))
    *   .catch((err) => console.log(err));
--- a/src/facilities/costCenters.js
+++ b/src/facilities/costCenters.js
@@ -1,0 +1,300 @@
+import isPlainObject from 'lodash.isplainobject';
+import {
+  formatCostCenterFromServer,
+  formatCostCenterToServer,
+  formatCostCenterFacilityFromServer
+} from '../utils/facilities';
+
+/**
+ * Module that provides access to cost centers, and helps manage
+ * the relationship between those cost centers and facilities
+ *
+ * @typicalname contxtSdk.facilities.costCenters
+ */
+/**
+ * @typedef {Object} CostCenter
+ * @param {string} createdAt ISO 8601 Extended Format date/time string
+ * @param {string} [description]
+ * @param {string} id UUID
+ * @param {string} name
+ * @param {string} organizationId UUID
+ * @param {string} updatedAt ISO 8601 Extended Format date/time string
+ */
+/**
+ * @typedef {Object} CostCenterFacility
+ * @param {string} costCenterId UUID
+ * @param {string} createdAt ISO 8601 Extended Format date/time string
+ * @param {number} facilityId
+ * @param {string} id UUID
+ * @param {string} updatedAt ISO 8601 Extended Format date/time string
+ */
+class CostCenters {
+  /**
+   * @param {Object} sdk An instance of the SDK so the module can communicate with other modules
+   * @param {Object} request An instance of the request module tied to this module's audience.
+   * @param {string} baseUrl The base URL provided by the parent module
+   */
+  constructor(sdk, request, baseUrl) {
+    this._baseUrl = baseUrl;
+    this._request = request;
+    this._sdk = sdk;
+  }
+
+  /**
+   * Adds a facility to a cost center
+   *
+   * API Endpoint: '/costcenters/:costCenterId/facility/:facilityId'
+   * Method: POST
+   *
+   * @param {string} costCenterId UUID corresponding with a cost center
+   * @param {number} facilityId The ID of a facility
+   *
+   * @returns {Promise}
+   * @fulfill {CostCenterFacility} Information about the new cost center facility relationship
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.facilities.costCenters
+   *   .addFacility('b3dbaae3-25dd-475b-80dc-66296630a8d0', 4)
+   *   .then((costCenter) => console.log(costCenter))
+   *   .catch((err) => console.log(err));
+   */
+  addFacility(costCenterId, facilityId) {
+    let errorMsg;
+
+    if (!costCenterId) {
+      errorMsg =
+        'A costCenterId is required to create a relationship between a cost center and a facility.';
+    } else if (!facilityId) {
+      errorMsg =
+        'A facilityId is required to create a relationship between a cost center and a facility.';
+    }
+
+    if (errorMsg) {
+      return Promise.reject(new Error(errorMsg));
+    }
+
+    return this._request
+      .post(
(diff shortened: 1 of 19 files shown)


src/facilities/costCenters.js
```diff
@@ -0,0 +1,300 @@
...
+   * contxtSdk.facilities.costCenters
+   *   .update({
+   *     description: 'US States of CT, MA, ME, NH, RI, VT',
+   *     name: 'New England, USA',
+   *   })
+   *   .then((costCenter) => console.log(costCenter))
+   *   .catch((err) => console.log(err));
+   */
+  update(costCenterId, update) {
+    if (!costCenterId) {
```
grettaface: Minor, [up top](https://github.com/ndustrialio/contxt-sdk-js/pull/34/files#diff-46620d5ea7e11fbc46891e5d4413fa20R228) we are creating a variable for the errors and rejecting in one spot, here we are rejecting on each conditional.

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+      let formatCostCenterFacilityFromServer;
+      let promise;
+      let rawCostCenterFacility;
+      let request;
+
+      beforeEach(function() {
+        expectedCostCenterFacility = fixture.build('costCenterFacility');
+        expectedFacilityId = expectedCostCenterFacility.facilityId;
+        expectedCostCenterId = expectedCostCenterFacility.costCenterId;
+        rawCostCenterFacility = fixture.build('costCenterFacility', {
```
greg5green: If you did this as `rawCostCenterFacility = fixture.build('costCenterFacility', expectedCostCenterFacility, { fromServer: true });`, it should be the fixture to have the raw object to have the same values as the expected object (just with the different keys).

(This doesn't need to happen, just kind of an FYI and cool thing Rosie can do)

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+      let formatCostCenterToServer;
+      let formattedCostCenterFromServer;
+      let formattedCostCenterToServer;
+      let initialCostCenter;
+      let promise;
+      let request;
+
+      beforeEach(function() {
+        initialCostCenter = fixture.build('costCenter');
+        formattedCostCenterToServer = fixture.build('costCenter', null, {
```
greg5green: You don't need `fromServer` on this one.

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+
+        promise = costCenters.create(initialCostCenter);
+      });
+
+      it('formats the submitted cost center object to send to the server', function() {
+        expect(formatCostCenterToServer).to.be.calledWith(initialCostCenter);
+      });
+
+      it('creates a new cost center', function() {
+        expect(request.post).to.be.deep.calledWith(
```
greg5green: You shouldn't need `deep` here.

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+      });
+
+      it('returns a fulfilled promise', function() {
+        return expect(promise).to.be.fulfilled;
+      });
+    });
+
+    context('when there is missing required information', function() {
+      it(`it throws an error when the cost center id is missing`, function() {
+        const expectedErrorMessage = `A cost center id is required for deleting a cost center.`;
```
greg5green: Would be preferable to have regular, single quotes here.

src/facilities/costCenters.spec.js
greg5green: BUT, this _*DOES*_ need `null` passed in though for it to generate the right format

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+    beforeEach(function() {
+      const numberOfCostCenters = faker.random.number({
+        min: 1,
+        max: 10
+      });
+      expectedCostCenter = fixture.buildList('costCenter', numberOfCostCenters);
+      costCentersFromServer = fixture.buildList(
+        'costCenter',
+        numberOfCostCenters
+      );
```
greg5green: Should do `costCentersFromServer = fixture.buildList('costCenter', numberOfCostCenters, null, { fromServer: true });` (I think that will work, but my brain is still a little out of it from being off work)

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+    let costCentersFromServer;
+    let promise;
+    let request;
+
+    beforeEach(function() {
+      const numberOfCostCenters = faker.random.number({
+        min: 1,
+        max: 10
+      });
+      expectedCostCenter = fixture.buildList('costCenter', numberOfCostCenters);
```
greg5green: This should be `expectedCostCenters`

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+      let costCentersFromServer;
+      let promise;
+      let request;
+
+      beforeEach(function() {
+        const numberOfCostCenters = faker.random.number({
+          min: 1,
+          max: 10
+        });
+        expectedCostCenter = fixture.buildList(
```
greg5green: `expectedCostCenters` here too

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+          'costCenter',
+          numberOfCostCenters,
+          false
```
greg5green: What's this for?

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+        );
+        costCentersFromServer = fixture.buildList(
+          'costCenter',
+          numberOfCostCenters,
+          true
+        );
```
greg5green: I'm thinking this should be `costCentersFromServer = fixture.buildList('costCenter', numberOfCostCenters', null, { fromServer: true });`

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,540 @@
...
+        expectedOrganizationId = fixture.build('organization').id;
```
greg5green: If you generate this first, you can pass it in when building the cost centers, like `costCentersFromServer = fixture.buildList('costCenter', numberOfCostCenters', { organizationId: expectedOrganizationId }, { fromServer: true });

src/utils/facilities/formatCostCenterFromServer.spec.js
```diff
@@ -0,0 +1,29 @@
...
+      ['created_at', 'organization_id', 'updated_at']
+    );
+
+    formattedCostCenter = formatCostCenterFromServer(costCenter);
+  });
+
+  it('converts the object keys to camelCase', function() {
+    expect(formattedCostCenter).to.deep.equal(expectedCostCenter);
+  });
+});
```
greg5green: There should be a test about formatting children properties (like `facilities`). I think there is an example in: https://github.com/ndustrialio/contxt-sdk-js/blob/master/src/utils/facilities/formatFacilityFromServer.spec.js

src/utils/facilities/formatCostCenterToServer.spec.js
```diff
@@ -0,0 +1,25 @@
...
+        ...costCenter,
+        organization_id: costCenter.organizationId
+      },
+      ['createdAt', 'id', 'facilities', 'organizationId', 'updatedAt']
+    );
+
+    formattedCostCenter = formatCostCenterToServer(costCenter);
+  });
+
+  it('converts the object keys to snake case and capitalizes certain keys', function() {
```
greg5green: This description isn't quite right since it is not capitalizing anything

greg5green:
One other thing -- you still need to add `this.costCenters = new CostCenters(sdk, request, baseUrl);` to `src/facilities/index.js`

src/facilities/costCenters.js
jongamble: On this one, we are checking two things on the `update` with different error messages. If the first is not true, the second is also not true, but overwriting the error message.

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,548 @@
...
+      let formatCostCenterToServer;
+      let formattedCostCenterFromServer;
+      let formattedCostCenterToServer;
+      let initialCostCenter;
+      let promise;
+      let request;
+
+      beforeEach(function() {
+        initialCostCenter = fixture.build('costCenter');
+        formattedCostCenterToServer = fixture.build('costCenter', null);
```
greg5green: Minor quibble, but can you get rid of `null` here? (Will be clearer in the future instead of someone looking at this and wondering why it is needed)

src/facilities/costCenters.spec.js
```diff
@@ -0,0 +1,548 @@
...
+        );
+      });
+
+      it('returns a fulfilled promise', function() {
+        return expect(promise).to.be.fulfilled;
+      });
+    });
+
+    context('when there is missing required information', function() {
+      it(`it throws an error when the cost center id is missing`, function() {
```
greg5green: I should have noticed this before, but there is one too many its here and should have single quotes

src/utils/facilities/formatCostCenterFromServer.spec.js
jongamble: https://github.com/ndustrialio/contxt-sdk-js/pull/34/files#diff-31f28e295fbcb24bd72f174ad08447c0R39

Is that what you're looking for?

greg5green: Yes 👍

PR #35, opened 2018-06-14
Add ability to attach custom Axios interceptors

greg5green:
## Why?
It is important that requests and responses can be handled in custom ways (e.g. to automatically call a custom error handler on every response).

## What changed?
- Added a way to add custom Axios interceptors to the Request module

### DIFF
--- a/docs/Typedefs.md
+++ b/docs/Typedefs.md
@@ -24,6 +24,20 @@ A single audience used for authenticating and communicating with an individual A
 | apiToken    | <code>string</code> |
 | expiresAt   | <code>number</code> |
 
+<a name="AxiosInterceptor"></a>
+
+## AxiosInterceptor : <code>Object</code>
+
+An object of interceptors that get called on every request or response.
+More information at [axios Interceptors](https://github.com/axios/axios#interceptors)
+
+**Kind**: global typedef
+
+| Param                 | Type                  | Description                                                    |
+| --------------------- | --------------------- | -------------------------------------------------------------- |
+| interceptor.fulfilled | <code>function</code> | A function that is run on every successful request or response |
+| interceptor.rejected  | <code>function</code> | A function that is run on every failed request or response     |
+
 <a name="CostCenter"></a>
 
 ## CostCenter : <code>Object</code>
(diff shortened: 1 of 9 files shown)


docs/Typedefs.md
```diff
@@ -24,6 +24,20 @@ A single audience used for authenticating and communicating with an individual A
...
+An object of interceptors that get called on every request or response.
+More information at [axios Interceptors](https://github.com/axios/axios#interceptors)
+
+**Kind**: global typedef
+
+| Param                 | Type                  | Description                                                    |
+| --------------------- | --------------------- | -------------------------------------------------------------- |
+| interceptor.fulfilled | <code>function</code> | A function that is run on every successful request or response |
+| interceptor.rejected  | <code>function</code> | A function that is run on every failed request or response     |
+
```
leeallen337: This is probably me not understanding the interceptors that well since I just skimmed the axios interceptor section in their docs but it seemed to me that the interceptors are bifurcated between requests and responses and not fulled and rejected? Hope my question makes sense. Like for the `rejected`, how would the interceptor know that the request was reject since it hasn't received a response yet?

src/config/index.js
```diff
@@ -40,6 +40,16 @@ import defaultConfigs from './defaults';
...
+/**
+ * An object of interceptors that get called on every request or response.
+ * More information at {@link https://github.com/axios/axios#interceptors axios Interceptors}
+ *
+ * @typedef {Object} AxiosInterceptor
+ * @param {function} interceptor.fulfilled A function that is run on every successful request or
+ *   response
+ * @param {function} interceptor.rejected A function that is run on every failed request or response
+ */
+
```
leeallen337: Similar question about this being `interceptor.request` and `interceptor.response`

src/request.spec.js
```diff
@@ -105,6 +119,73 @@ describe('Request', function() {
...
+            rejected: new Function()
+          };
+        }
+      );
+      const responseInterceptors = times(
+        faker.random.number({ min: 0, max: 10 }),
+        () => {
+          return {
+            fulfilled: new Function(),
+            rejected: new Function()
```
leeallen337: Could you use `this.sandbox.stub()` here instead of `new Function()`?

src/request.js
```diff
@@ -8,10 +8,11 @@ class Request {
    */
   constructor(sdk, audienceName) {
     this._audienceName = audienceName;
-    this._sdk = sdk;
     this._axios = axios.create();
+    this._insertHeaders = this._insertHeaders.bind(this);
+    this._sdk = sdk;
 
-    this._axios.interceptors.request.use(this._insertHeaders);
+    this._attachInterceptors(this._axios);
```
leeallen337: Is it needed to pass `this._axios` here? It appears that the `_attachInterceptors` function declaration below doesn't take any arguments?

src/request.spec.js
greg5green: Yes, for sure. I had a brain fart trying to think of some method in faker. Somehow didn't make it to sinon.

src/request.js
greg5green: 👍 Vestigal code smh

docs/Typedefs.md
greg5green: For the words here, I went with the verbiage axios uses in their code (which doesn't really differentiate between request or response: https://github.com/axios/axios/blob/master/lib/core/InterceptorManager.js#L17).

In the case of rejected requests, I believe those interceptors would be invoked if there was an error in a previous interceptor (and would still return .a promise)

PR #36, opened 2018-06-19
I didn't bump the version to 0.0.14

leeallen337:
## Why?

Because I forgot, when I rebased, to bump the version to `0.0.14`

## What changed?
Updated the version in `package.json`


rob5standingby:
:shipit:

PR #39, opened 2018-06-28
Add asset_id to the facility model

jrcrawfo:
## Why?
- Need to be able to retrieve the asset reference from the facility model (recently added to the facilities API)

## What changed?
- Added the asset_id field to the facility model

## What's to Come?
- This serves a short term need to just retrieve the asset_id, but we need to make an effort to build the asset framework into this library in the near future.

### DIFF
--- a/src/facilities/index.js
+++ b/src/facilities/index.js
@@ -11,6 +11,7 @@ import {
  * @typedef {Object} Facility
  * @property {string} [address1]
  * @property {string} [address2]
+ * @property {string} [assetId] UUID corresponding with an asset
  * @property {string} [city]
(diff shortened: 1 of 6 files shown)


greg5green:
1. This should be added to the fixture factory (https://github.com/ndustrialio/contxt-sdk-js/blob/master/support/fixtures/factories/facility.js#L14)
2. The new field is documented as optional in a few places, but not all -- should it be documented as optional everywhere?
3. This is really nitpicky, but everything was sorted alphabetically before -- can you move `assetId`/`asset_id` so that this trend continues?

greg5green:
:shipit:

PR #42, opened 2018-07-01
Fixes issue with machineAuth getSessionInfo for multiple audiences

rob5standingby:
## Why?
- Small bug with using `machineAuth` that I discovered while working on EMS API. When calling two functions from two different modules at the same time (for the first time), the incorrect token would get utilized for the second function call.

## What changed?
- Adjusted the function `getSessionInfo` to return a hash of promises, rather than a single promise

### DIFF
--- a/src/sessionTypes/machineAuth.js
+++ b/src/sessionTypes/machineAuth.js
@@ -101,8 +101,12 @@ class MachineAuth {
       return Promise.reject(new Error('No valid audience found'));
     }
 
-    if (!this._tokenPromise) {
-      this._tokenPromise = axios
+    if (!this.tokenPromises) {
+      this.tokenPromises = {};
+    }
+
+    if (!this.tokenPromises[audienceName]) {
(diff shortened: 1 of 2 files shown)


src/sessionTypes/machineAuth.js
```diff
@@ -101,8 +101,12 @@ class MachineAuth {
       return Promise.reject(new Error('No valid audience found'));
     }
 
-    if (!this._tokenPromise) {
-      this._tokenPromise = axios
+    if (!this.tokenPromises) {
+      this.tokenPromises = {};
```
greg5green: Maybe we should do this in the constructor?

rob5standingby: Yeah good thinking.

rob5standingby:
@greg5green I've moved instantiation of `this.tokenPromises` to the constructor, and also renamed the var to `this._tokenPromises` to stick with the naming conventions. Let me know if this looks good!

PR #53, opened 2018-07-20
Allow global asset type creation

greg5green:
## Why?
To create a global asset type, `null` is explicitly passed for the organization ID. We were doing a falsey check to make sure the required fields were present, which failed if passing in `null`.

## What changed?
- Wrote custom (and default) "check" functions mechanism to allow for checking if a field is present in different ways

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,6 @@
 
 * Updated AssetTypes#create to allow for the creation of global asset types
 
-**Changed**
-
-* Now supporting the `data_type` field for `AssetAttributes`
-
 ## [v0.0.20](http://github.com/ndustrialio/contxt-sdk-js/tree/v0.0.20) (2018-07-16)
 
 **Added**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## [v0.0.20](http://github.com/ndustrialio/contxt-sdk-js/tree/v0.0.21) (2018-07-23)
+
+**Fixed**
+
+* Updated AssetTypes#create to allow for the creation of global asset types
+
+**Changed**
+
+* Now supporting the `data_type` field for `AssetAttributes`
+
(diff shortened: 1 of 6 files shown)


src/assets/assetTypes.js
```diff
@@ -66,12 +68,17 @@ class AssetTypes {
...
+      organizationId: (object, key) => has(object, key)
+    };
     const requiredFields = ['description', 'label', 'organizationId'];
 
     for (let i = 0; i < requiredFields.length; i++) {
       const field = requiredFields[i];
+      const hasField = hasFieldFns[field] || hasFieldFns.default;
 
-      if (!assetType[field]) {
+      if (!hasField(assetType, field)) {
```
leeallen337: Instead of having the `hasFieldFns` object and the `hasField` constant, couldn't you just do the following inside the `if` check here?
```
if (!has(assetType, field)) {
```

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## [v0.0.20](http://github.com/ndustrialio/contxt-sdk-js/tree/v0.0.21) (2018-07-23)
+
+**Fixed**
+
+* Updated AssetTypes#create to allow for the creation of global asset types
+
+**Changed**
+
+* Now supporting the `data_type` field for `AssetAttributes`
```
leeallen337: Is this in another pull request? I don't see any files changed that include `data_type`?

greg5green: I did a bad copy and paste job. Will remove.

src/assets/assetTypes.js
greg5green: Not really -- `organization_id` is the only field that is allowed to be null, the others require a non-null and non-undefined value. (`_.has` returns true if `undefined` or `null` is explicitly set on a property)

greg5green: If I've misunderstood what you're asking, lmk and I'll change it in another PR

leeallen337: Oh my mistake. For some reason I thought `_.has` returns `true` as long as the value related to the key is not `undefined` so `null` would be okay

PR #54, opened 2018-07-26
Update Asset Attribute Value create endpoint

leeallen337:
## Why?

https://ndustrialio.atlassian.net/browse/FM-138

The current API endpoint for creating an asset attribute value is too generic and should be updated since we will start supporting values for asset metrics

## What changed?
- Changed API endpoint
- Updated tests
- Updated docs

#### NOTE: This pull request is tied to https://github.com/sustainableis/facilities-service/pull/50


rob5standingby:
@leeallen337 This looks great but looks like this branch needs a bit of updating now that `master` has a few new versions since this PR was first created

leeallen337:
@rob5standingby Updated it 👍

PR #57, opened 2018-08-06
Update attribute value to reflect model changes

jongamble:
## Why?
adding assetLabel and label to the attribute value to reflect new changes to attribute value model

## What changed?
adding `label` and `assetLabel` to the attribute value model

requires sustainableis/facilities-service#51

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ndustrial/contxt-sdk",
-  "version": "0.0.23",
+  "version": "0.0.24",
   "description": "",
   "main": "lib/index.js",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ndustrial/contxt-sdk",
-  "version": "0.0.23",
+  "version": "0.0.24",
```
greg5green: This should have happened after a CHANGELOG update

jongamble: dammit. I knew I was forgetting something.

PR #66, opened 2018-08-15
Switch to generic/non-bespoke formatters

### LINKED ISSUE #47: Update formatter utils to not be bespoke for each data type
When starting the project, I decided to make each of the formatter utils be bespoke for each data type to deal with some edge cases related to facilities and facility info. A few months into the project, it's become obvious that this has introduced technical debt for us -- it is turning out to be significantly more work to make two new formatters for each data type (to and from the server).

We should re-engineer this so that most cases use one formatter for each direction (to the server and from the server) and then have a couple separate formatters to handle edge cases.

greg5green:
## Why?
1. Early on, it was a [poor] design choice to create a formatter for each data type. It was getting unmaintainable and hard to manage. (See #47)
2. With update actions, we were a little too permissive with what was getting passed through -- if passing in the original object returned from an API with only one field updated, it would usually throw an error since the APIs were checking to make sure certain keys were not being updated.

## What changed?
- Added `toCamelCase` and `toSnakeCase` generic formatters
  - Utilizes [`change-case`](https://github.com/blakeembrey/change-case) to manage changing the case of keys
  - Uses a `mapObj` function which is based on the [`map-obj` library](https://github.com/sindresorhus/map-obj), with the added ability to leave/exclude keys from the new object
  - `createCaseChangeFn` function is used to create different "change case" functions. It's only used to create ones for snake_case and camelCase at the moment, but can be used to create any case provided in `change-case` (or any custom functions that operate the same as functions provided in `change-case`)
  - Based on [`camelcase-keys`](https://github.com/sindresorhus/camelcase-keys), but with the ability to exclude keys and convert to other cases than just camelCase

Closes #47


src/utils/objects/map.js
```diff
@@ -0,0 +1,34 @@
+import isPlainObject from 'lodash.isplainobject';
+
+function map(input, callback, options) {
```
leeallen337: I'm assuming you put `options` as the last argument since you may not have `options` to pass down. I'm not sure how other libraries do this but it does feel a little odd to me to not have the iterator function last but I'm not sure I can think of a better solution where the iterator is always last

src/utils/objects/toCamelCase.spec.js
```diff
@@ -0,0 +1,203 @@
...
+        second_key: 'second value'
+      });
+    });
+
+    it('does not transform nested excluded keys', function() {
+      const transformedObject = toCamelCase(
+        {
+          first_level_first_key: {
+            second_level_first_key: {
+              third_levl_first_key: 'third level first value'
```
leeallen337: `third_level_first_key`?

src/utils/pagination/formatPaginatedDataFromServer.js
```diff
@@ -9,7 +11,8 @@
  * @private
  */
 function defaultFormatter(value) {
-  return value;
+  // debugger;
```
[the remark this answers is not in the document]
greg5green: 👍

src/utils/objects/map.js
```diff
@@ -0,0 +1,34 @@
...
+
+    if (options.deep) {
+      if (isPlainObject(newValue)) {
+        newValue = map(newValue, callback, options);
+      } else if (Array.isArray(newValue)) {
+        newValue = newValue.map(
+          (val) => (isPlainObject(val) ? map(val, callback, options) : val)
+        );
+      }
+    }
```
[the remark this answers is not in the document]
greg5green: 1. I believe the answer is yes.
2. Can you provide an example object of what you mean specifically?

src/utils/facilities/formatFacilityWithInfoFromServer.js
```diff
@@ -52,46 +47,16 @@ import {
...
-  if (input.cost_centers) {
-    facility.costCenters = input.cost_centers.map(formatCostCenterFromServer);
-  }
-
-  return facility;
+  return formattedFacility;
 }
 
-export default formatFacilityFromServer;
+export default formatFacilityWithInfoFromServer;
```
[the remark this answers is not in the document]
greg5green: And that the data in it shouldn't be converted either since they are custom keys. It does a mapping when sending it from `[{ key_1: 'value 1' }, { key_2: 'value 2' }]` to `{ key_1: 'value 1', key_2: 'value 2' }`

src/utils/objects/toCamelCase.spec.js
greg5green: 👍

src/utils/objects/map.js
greg5green: The native APIs for stuff like `map` and `forEach` do this (optional arguments last, specifically the `thisArg`), as does Underscore. It's pretty common.

leeallen337: What about doing something like the following?
```
function map(input, options, callback) {
  callback = callback || options;
  options = typeof options === 'function'
    ? { deep: false }
    : { deep: false, ...options }

  // the rest of the function here
}
```

leeallen337: Also should there be any JSDocs for these private functions?

src/assets/assetAttributes.spec.js
```diff
@@ -732,9 +723,7 @@ describe('Assets/Attributes', function() {
 
       it('formats the asset attribute values', function() {
         return promise.then(() => {
-          valuesFromServerBeforeFormat.forEach((value) => {
-            expect(formatAssetAttributeValueFromServer).to.be.calledWith(value);
-          });
+          expect(toCamelCase).to.be.calledWith(valuesFromServerBeforeFormat);
```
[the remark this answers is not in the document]
greg5green: 1. Yeah, there is no `deep.calledWith` in sinon-chai.
2. I'm going to remove those `deep.calledWith`s. I didn't necessarily parse out that syntax when switching the method names used.

src/utils/objects/createCaseChangeFn.js
```diff
@@ -0,0 +1,36 @@
+import mapObj from './map';
+
+function createCaseChangeFn(caseChangeFn) {
+  function changeCase(input = {}, options) {
+    options = {
+      deep: true,
+      excludeKeys: [],
+      excludeTransform: [],
+      ...options
+    };
```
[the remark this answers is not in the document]
greg5green: Yeah, I should have a separate variable. I'll change.

src/utils/objects/map.js
greg5green: 1. All evidence I have seen (how native array iterators and Underscore handle optional args, [AirBnb style guide on default parameters](https://github.com/airbnb/javascript#functions--defaults-last)) points to your assumption (the iterator needs to be last in this case) being wrong. I think it all stems back to function overloading not being something built into JS. Are you aware of any counter-examples off the top of your head? I know I've seen them before. I'd really rather avoid doing two reassignments to just move the optional argument.

1. Yes, there should be JSDocs for these. Whoops.

src/utils/objects/map.js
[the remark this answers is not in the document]
greg5green: The answer is yes. It's an easy test, so I'll add it

PR #72, opened 2018-09-24
Feature/event methods

grettaface:
## Why?
- Needed to add endpoints for getting events by type and getting event types by client

## What changed?
- Integrated those endpoints into the SDK
- Added tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ndustrial/contxt-sdk",
-  "version": "0.0.32",
+  "version": "0.0.33",
   "description": "",
   "main": "lib/index.js",
   "module": "es/index.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ndustrial/contxt-sdk",
-  "version": "0.0.33",
+  "version": "0.0.32",
   "description": "",
   "main": "lib/index.js",
   "module": "es/index.js",
--- a/src/events/index.js
+++ b/src/events/index.js
@@ -1,6 +1,7 @@
 import isPlainObject from 'lodash.isplainobject';
 import { formatEventUpdateToServer } from '../utils/events';
 import { toCamelCase, toSnakeCase } from '../utils/objects';
+import { formatPaginatedDataFromServer } from '../utils/pagination';
 
 /**
  * @typedef {Object} Event
@@ -34,6 +35,35 @@ import { toCamelCase, toSnakeCase } from '../utils/objects';
  * @property {string} [ownerId] The ID of the user who owns the event
  * @property {number} [topicArn] The Amazon Resource Name (ARN) associated with the event
  * @property {string} updatedAt ISO 8601 Extended Format date/time string
+ */
+
+/**
+ * @typedef {Object} EventType
+ * @property {string} client_id UUID corresponding with the client
+ * @property {string} created_at ISO 8601 Extended Format date/time string
+ * @property {string} description
+ * @property {string} id UUID
+ * @property {boolean} is_realtime_enabled Specifies whether that event types is enabled for realtime
+ * @property {number} level Priority level associated with event type
+ * @property {string} name
+ * @property {string} slug
+ * @property {string} updatedAt ISO 8601 Extended Format date/time string
+ */
+
+/**
+ * @typedef {Object} EventTypesFromServer
+ * @property {Object} _metadata Metadata about the pagination settings
+ * @property {number} _metadata.offset Offset of records in subsequent queries
+ * @property {number} _metadata.totalRecords Total number of asset types found
+ * @property {EventType[]} records
+ */
+
+/**
+ * @typedef {Object} EventsFromServer
+ * @property {Object} _metadata Metadata about the pagination settings
+ * @property {number} _metadata.offset Offset of records in subsequent queries
+ * @property {number} _metadata.totalRecords Total number of asset types found
+ * @property {Events[]} records
  */
 
 /**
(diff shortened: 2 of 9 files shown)


src/events/index.js
```diff
@@ -160,6 +160,74 @@ class Events {
...
+  /**
+   * Gets all event types for a client
+   *
+   * API Endpoint: '/clients/:clientId/types'
+   * Method: GET
+   *
+   * @param {string} clientID The ID of the client
+   *
+   * @returns {Promise}
+   * @fulfill {Event} Information about an event
```
leeallen337: Is this returning a single event or all event types related to a client? On line 164 it looks like it's returning all `EventTypes` related to a specific client but this reads like it's returning a single event?

src/events/index.js
```diff
@@ -160,6 +160,74 @@ class Events {
...
+    }
+    return this._request.get(`${this._baseUrl}/clients/${clientId}/types`);
+  }
+  /**
+   * Gets all events by type
+   *
+   * API Endpoint: '/types/:typeId/events?facility_id=:facility_id&include[]=latest'
+   * Method: GET
+   *
+   * @param {string} id The ID of the type
```
leeallen337: Personal preference here but since there are multiple id's here I think it would be more clear if this was changed to `eventTypeId` or something similar

src/events/index.js
```diff
@@ -160,6 +160,74 @@ class Events {
...
+        new Error('A type id is required for getting events')
+      );
+    }
+    if (!facilityId) {
+      return Promise.reject(
+        new Error('A facility id is required for getting events')
+      );
+    }
+    const params = {
+      facility_id: facilityId
```
greg5green: We shouldn't be doing this manually -- should be using `toSnakeCase`, like:
```
const params = { facilityId };

if (latest) {
  params.include = ['triggered.latest'];
}

return this.request.get(`${this._baseUrl}/types/${id}/events`, toSnakeCase({ params })
  .then((events) => toCamelCase(events));
```

src/events/index.spec.js
```diff
@@ -239,6 +239,113 @@ describe('Events', function() {
...
+
+    context('all required params are passed', function() {
+      it('gets the events from the server', function() {
+        expect(request.get).to.be.calledWith(
+          `${expectedHost}/types/${typeId}/events`
+        );
+      });
+
+      it('returns the requested events', function() {
+        return expect(promise).to.be.fulfilled.and.to.eventually.deep.equal(
```
greg5green: I don't think you need deep here

src/events/index.spec.js
```diff
@@ -239,6 +239,113 @@ describe('Events', function() {
...
+      });
+
+      it('gets the eventType from the server', function() {
+        expect(request.get).to.be.calledWith(
+          `${expectedHost}/clients/${clientId}/types`
+        );
+      });
+
+      it('returns the requested event type', function() {
+        return expect(promise).to.be.fulfilled.and.to.eventually.deep.equal(
```
greg5green: Shouldn't need deep here

src/events/index.spec.js
```diff
@@ -239,6 +239,113 @@ describe('Events', function() {
     });
   });
 
+  describe('getEventTypesByClient', function() {
+    context('the clientId is provided', function() {
+      let clientId = faker.random.uuid();
```
greg5green: The creation of `clientId` should happen in the before each -- and then pass it into `buildList` so you get the same clientId in all the fixtures

src/events/index.spec.js
```diff
@@ -239,6 +239,113 @@ describe('Events', function() {
...
+      let promise;
+      let request;
+      let eventTypeFromServer;
+
+      beforeEach(function() {
+        eventTypeFromServer = fixture.build('eventType', { fromServer: true });
```
greg5green: I believe you need a `null` in here before `fromServer`

src/events/index.js
```diff
@@ -160,6 +160,74 @@ class Events {
...
+    };
+
+    if (latest) {
+      params.include = ['triggered.latest'];
+    }
+
+    return this._request.get(`${this._baseUrl}/types/${id}/events`, { params });
```
leeallen337: I think both of these end points return paginated results which might need extra work. See https://github.com/ndustrialio/contxt-sdk-js/blob/master/src/assets/assetMetrics.js#L4
how we're handling some paginated data.

src/events/index.spec.js
```diff
@@ -239,6 +239,113 @@ describe('Events', function() {
...
+  describe('getEventsByType', function() {
+    let typeId = faker.random.uuid();
+    let facilityId = faker.random.uuid();
+    let promise;
+    let request;
+    let eventsFromServer;
+
+    beforeEach(function() {
+      eventsFromServer = times(
+        fixture.build('event', { fromServer: true }),
```
greg5green: `eventsFromServer = fixture.buildList('event', 10, null, { fromServer: true });`

(even better, use `faker.random.number({ min: 5, max: 20 })` instead of `10`)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ndustrial/contxt-sdk",
-  "version": "0.0.32",
+  "version": "0.0.33",
```
leeallen337: Regarding updating the numbers here and in the lock file. I'm assuming you just did that in an editor and didn't run `npm version x.x.x` or whatever. Take a look at the [Contributing and Publishing](https://github.com/ndustrialio/contxt-sdk-js#contributingpublishing) section in the README file. We do it a little differently so these shouldn't actually be updated like this.

src/events/index.js
```diff
@@ -160,6 +160,74 @@ class Events {
...
+   */
+  getEventsByType(id, facilityId, latest = false) {
+    if (!id) {
+      return Promise.reject(
+        new Error('A type id is required for getting events')
+      );
+    }
+    if (!facilityId) {
+      return Promise.reject(
+        new Error('A facility id is required for getting events')
```
greg5green: `ID` should be capitalized in the error messages

src/events/index.spec.js
greg5green: Also, shouldn't this be a list?

package.json
grettaface: Yeah, Greg mentioned that as well. Will do going forward.

src/events/index.js
```diff
@@ -160,6 +160,74 @@ class Events {
...
+   * @fulfill {Event} Information about an event
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.events
+   *   .getEventTypesByClient('CW4B1Ih6M1nNwwxk0XOKI21MVH04pGUL')
+   *   .then((event) => console.log(event))
+   *   .catch((err) => console.log(err));
+   */
+  getEventTypesByClient(clientId) {
```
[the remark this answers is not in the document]
grettaface: :+1:

src/events/index.spec.js
grettaface: They dont seem to pass w/o `.deep` which is why I added it originally

src/events/index.spec.js
```diff
@@ -185,19 +187,17 @@ describe('Events', function() {
 
       beforeEach(function() {
         expectedEventId = faker.random.uuid();
-        eventFromServerAfterFormat = fixture.build('event', {
-          id: expectedEventId
+        eventFromServerAfterFormat = fixture.build('event', expectedEventId);
```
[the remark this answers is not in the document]
grettaface: You're right. I was cleaning stuff up and think this one got changed as part of it. Reverting

src/events/index.js
```diff
@@ -160,6 +190,82 @@ class Events {
...
+      return Promise.reject(
+        new Error('A client ID is required for getting types')
+      );
+    }
+
+    return this._request
+      .get(`${this._baseUrl}/clients/${clientId}/types`)
+      .then((response) => formatPaginatedDataFromServer(response));
+  }
+  /**
```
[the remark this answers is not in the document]
grettaface: Not sure why this is still yelling. I see the line on mine and have no changes to commit. 
image

src/events/index.spec.js
greg5green: Oh, I see why. You need to stub out `formatPaginatedDataFromServer` and test that it is called

src/events/index.spec.js
```diff
@@ -239,6 +243,168 @@ describe('Events', function() {
...
+      it('formats the events object', function() {
+        return promise.then(() => {
+          expect(formatPaginatedDataFromServer).to.be.calledWith(
+            eventsFromServerBeforeFormat
+          );
+        });
+      });
+
+      it('returns the requested events', function() {
+        return expect(promise).to.be.fulfilled.and.to.eventually.deep.equal(
```
greg5green: `formatPaginatedDataFromServer` is stubbed here, so don't need deep here

grettaface:
@greg5green I was just intentionally leaving things out so we could keep interfacing with each other.

PR #75, opened 2018-10-11
bugfix/remove_organizationId_from_asset_attribute_required

erichardson30:
## Why?
Unable to create a global asset type attribute

## What changed?
- Removed `organizationId` from the list of required fields.
- Updated tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,295 +1,301 @@
+## [v0.0.35](http://github.com/ndustrialio/contxt-sdk-js/tree/v0.0.35) (2018-10-11)
+
+**Fixed**
+
+- Updated `AssetAttributes#create` and removed `organizationId` from the required fields.
+
 ## [v0.0.34](http://github.com/ndustrialio/contxt-sdk-js/tree/v0.0.34) (2018-10-08)
 
 **Fixed**
 
-* There were some methods that returned paginated data, but did not pass the `limit` and `offset` to the API. This has been fixed and now allows for passing `limit` and `offset` for:
-  * `AssetTypes#getAll`
-  * `AssetTypes#getAllByOrganizationId`
-  * `Assets#getAll`
-  * `Events#getEventTypesByClientId`
+- There were some methods that returned paginated data, but did not pass the `limit` and `offset` to the API. This has been fixed and now allows for passing `limit` and `offset` for:
+  - `AssetTypes#getAll`
+  - `AssetTypes#getAllByOrganizationId`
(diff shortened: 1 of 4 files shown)


src/assets/assetAttributes.js
```diff
@@ -117,7 +117,7 @@ class AssetAttributes {
    *   .catch((err) => console.log(err));
    */
   create(assetTypeId, assetAttribute = {}) {
-    const requiredFields = ['description', 'label', 'organizationId'];
+    const requiredFields = ['description', 'label'];
```
leeallen337: So I took a look at the creation of Asset Type and Asset Metric to see how we were handling `null` there. We still require `organizationId` as a `requiredField` but we do it a little differently. See [Asset Metric](https://github.com/ndustrialio/contxt-sdk-js/blob/master/src/assets/assetMetrics.js#L102-L122) for reference and [Asset Type](https://github.com/ndustrialio/contxt-sdk-js/blob/master/src/assets/assetTypes.js#L68-L83) as well. 

Also there's an added test for creating with a global scope [here](https://github.com/ndustrialio/contxt-sdk-js/blob/master/src/assets/assetTypes.spec.js#L122) which should probably be added to this as well.

docs/AssetAttributes.md
```diff
@@ -50,7 +50,7 @@ Method: POST
 | assetAttribute.description | <code>string</code> |  |
 | [assetAttribute.isRequired] | <code>boolean</code> |  |
 | assetAttribute.label | <code>string</code> |  |
-| assetAttribute.organizationId | <code>string</code> |  |
+| assetAttribute.organizationId | <code>string</code> | Can be explicitly set to `null` to create a global attribute  |
```
leeallen337: These are automatically generated so it's best not to make the change here and instead make it in the JSDoc stuff in `src/assets/assetAttributes.js` above the `create` method [here](https://github.com/ndustrialio/contxt-sdk-js/blob/master/src/assets/assetAttributes.js#L99)

CHANGELOG.md
```diff
@@ -1,295 +1,301 @@
+## [v0.0.35](http://github.com/ndustrialio/contxt-sdk-js/tree/v0.0.35) (2018-10-11)
+
+**Fixed**
+
+- Updated `AssetAttributes#create` and removed `organizationId` from the required fields.
```
leeallen337: Might want to change this wording to something like 

Updated `AssetAttributes#create` to allow for the creation of global asset attributes.

PR #79, opened 2019-01-16
Adds files module and get function

rob5standingby:
## Why?
We need a `Files` module and a `get` function to retrieve some metadata about a file from the new Files API!

## JIRA
https://ndustrialio.atlassian.net/browse/CFILE-2 📖 

## What changed?
- Added `Files` module. Based the File typedef on [this](https://contxt.readme.io/reference#files-retrieve-a-file).
- Added factory for `file`
- Added tests

### DIFF
--- a/src/config/audiences.js
+++ b/src/config/audiences.js
@@ -51,11 +51,11 @@ export default {
   },
   files: {
     production: {
-      clientId: '6jPAqJCMFMRP5t7Rw2X19qt2EtTfdQ2Q',
+      clientId: 'VZJ2MRd71dM9oX2hB1EVAEjUe6loL7pl',
       host: 'https://files.api.ndustrial.io'
     },
     staging: {
-      clientId: '6jPAqJCMFMRP5t7Rw2X19qt2EtTfdQ2Q',
+      clientId: 'SLE310LY4nDT5p06sUGPVOLlGXk5osbQ',
       host: 'https://files-staging.api.ndustrial.io'
     }
   },
--- a/src/config/audiences.js
+++ b/src/config/audiences.js
@@ -49,6 +49,16 @@ export default {
       host: 'https://facilities-staging.api.ndustrial.io'
     }
   },
+  files: {
+    production: {
+      clientId: '6jPAqJCMFMRP5t7Rw2X19qt2EtTfdQ2Q',
+      host: 'https://files.api.ndustrial.io'
+    },
+    staging: {
+      clientId: '6jPAqJCMFMRP5t7Rw2X19qt2EtTfdQ2Q',
+      host: 'https://files-staging.api.ndustrial.io'
+    }
+  },
   iot: {
     staging: {
       clientId: 'm35AEcxD8hf65sq04ZU7yFxqpqVkKzES',
--- a/src/files/index.js
+++ b/src/files/index.js
@@ -1,0 +1,65 @@
+import { toCamelCase } from '../utils/objects';
+
+/**
+ * @typedef {Object} File
+ * @property {string} createdAt ISO 8601 Extended Format date/time string
+ * @property {string} contentType The MIME type of the file
+ * @property {string} description
+ * @property {string} filename
+ * @property {number} id
+ * @property {string} organizationId UUID of the organization to which the file belongs
+ * @property {string} ownerId The ID of the user who owns the file
+ * @property {string} status The status of the File, e.g. "ACTIVE"
+ * @property {string} updatedAt ISO 8601 Extended Format date/time string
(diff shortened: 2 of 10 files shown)


src/files/index.js
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @param {string} fileId The ID of the file
+   *
+   * @returns {Promise}
+   * @fulfill {File} Information about a file
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.files
+   *   .get(12345)
```
leeallen337: Aren't the id's UUIDs?

src/files/index.js
```diff
@@ -0,0 +1,65 @@
...
+   *   .then((file) => console.log(file))
+   *   .catch((err) => console.log(err));
+   */
+  get(fileId) {
+    if (!fileId) {
+      return Promise.reject(
+        new Error('A file ID is required for getting information about a file')
```
leeallen337: Do we put periods at the end of our error messages or no?

src/files/index.js
leeallen337: https://github.com/ndustrialio/files-api/blob/master/models/file.js#L12

support/fixtures/factories/file.js
```diff
@@ -0,0 +1,35 @@
...
+
+factory
+  .define('file')
+  .option('fromServer', false)
+  .attrs({
+    createdAt: () => faker.date.past().toISOString(),
+    contentType: () => faker.system.mimeType(),
+    description: () => faker.hacker.phrase(),
+    filename: () => `${faker.hacker.noun()}.${faker.system.commonFileExt()}`,
+    id: () => faker.random.number(),
```
leeallen337: Looks like this should be a UUID.
https://github.com/ndustrialio/files-api/blob/master/models/file.js#L12

src/files/index.js
rob5standingby: Greg and I are out of sync. I documented them as Integers ;P

src/files/index.js
rob5standingby: No, it doesn't appear so based on some of the other modules

src/config/audiences.js
```diff
@@ -49,6 +49,16 @@ export default {
...
     }
   },
+  files: {
+    production: {
+      clientId: '6jPAqJCMFMRP5t7Rw2X19qt2EtTfdQ2Q',
+      host: 'https://files.api.ndustrial.io'
+    },
+    staging: {
+      clientId: '6jPAqJCMFMRP5t7Rw2X19qt2EtTfdQ2Q',
+      host: 'https://files-staging.api.ndustrial.io'
```
greg5green: These client IDs are different now. And might change again in the future for reasons.

Current:
Staging: `SLE310LY4nDT5p06sUGPVOLlGXk5osbQ`
Production: `VZJ2MRd71dM9oX2hB1EVAEjUe6loL7pl`

src/files/index.spec.js
```diff
@@ -0,0 +1,121 @@
...
+        expectedFileId = faker.random.uuid();
+        fileFromServerAfterFormat = fixture.build('file', {
+          id: expectedFileId
+        });
+
+        fileFromServerBeforeFormat = fixture.build(
+          'file',
+          { id: expectedFileId },
+          { fromServer: true }
+        );
```
greg5green: What do you think about doing this a little differently?

```
fileFromServerAfterFormat = fixture.build('file');
fileFromServerBeforeFormat = fixture.build('file', fileFromServerAfterFormat, { fromServer: true });
```

support/fixtures/factories/file.js
```diff
@@ -0,0 +1,35 @@
...
+const faker = require('faker');
+
+factory
+  .define('file')
+  .option('fromServer', false)
+  .attrs({
+    createdAt: () => faker.date.past().toISOString(),
+    contentType: () => faker.system.mimeType(),
+    description: () => faker.hacker.phrase(),
+    filename: () => `${faker.hacker.noun()}.${faker.system.commonFileExt()}`,
```
greg5green: What about `() => faker.system.commonFileName()`?

rob5standingby: Sure, wasn't aware of that one!

src/files/index.spec.js
rob5standingby: Ah, I keep forgetting about that option.

src/config/audiences.js
rob5standingby: Ah, thanks. I'll update them to those for now.

PR #81, opened 2019-01-24
PRO-282: Add the passwordGrantAuth module, allows authenticating using the "password" grant type for Auth0

leeallen337:
## Why?

https://ndustrialio.atlassian.net/browse/PRO-282

As a user of contxt, I want to be able to login through the command line

## What changed?
- Added a `cliAuth` SessionType to the contxt sdk
- Added a new module with some methods for logging in and logging out


src/sessionTypes/cliAuth.js
```diff
@@ -0,0 +1,153 @@
...
+          if (err) {
+            const errorMessage =
+              (err && err.description) || 'Authentication failed.';
+
+            return reject(new Error(errorMessage));
+          }
+
+          this._saveSession('accessInfo', response);
+
+          return resolve('Authentication successful.');
```
greg5green: What do you think of resolving with nothing here -- you're not using it in the `then` at all. Or resolving with the accessToken (I think I like this more)

src/sessionTypes/cliAuth.js
```diff
@@ -0,0 +1,153 @@
...
+        {
+          headers: { Authorization: `Bearer ${accessToken}` }
+        }
+      )
+      .then((response) => {
+        const { data } = response;
+
+        this._saveSession('apiToken', data);
+
+        return 'Contxt Authentication successful.';
```
greg5green: I'm not sure you need this message either. I think you should return the token to match what the method name is

src/sessionTypes/cliAuth.js
```diff
@@ -0,0 +1,153 @@
...
+ *   config: {
+ *     auth: {
+ *       clientId: 'bleED0RUwb7CJ9j7D48tqSiSZRZn29AV'
+ *     }
+ *   },
+ *   sessionType: 'cliAuth'
+ * });
+ */
+
+class CliAuth {
```
greg5green: You need a couple more methods here:
1. `getCurrentApiToken`
2. `isAuthenticated`

These are both [documented](https://github.com/ndustrialio/contxt-sdk-js/blob/276a9bbafbca78a5e8976019c7088bf9b69c1308/docs/Typedefs.md#sessiontype--object]) as methods that are required for a SessionType

src/sessionTypes/cliAuth.js
```diff
@@ -0,0 +1,153 @@
...
+   * Logs the user out by removing any stored session info.
+   *
+   * @returns {Promise}
+   * @fulfills {string}
+   */
+  logOut() {
+    return new Promise((resolve) => {
+      this._sessionInfo = {};
+
+      return resolve('Logout successful - session info cleared.');
```
greg5green: 1. What's the thought process behind making this async?
1. This one doesn't bother me as much because there's nothing obvious that should be returned from a method like this, but I don't see the utility in this really.

src/sessionTypes/cliAuth.spec.js
```diff
@@ -0,0 +1,394 @@
...
+          audiences: {
+            contxtAuth: fixture.build('audience'),
+            facilities: fixture.build('audience')
+          },
+          auth: {
+            clientId: faker.internet.password()
+          }
+        }
+      };
+      authenticationSession = {
```
greg5green: You should be able to leave this stub out here as it's not being used

src/sessionTypes/cliAuth.spec.js
```diff
@@ -0,0 +1,394 @@
...
+        return promise.then(() => {
+          expect(cliAuth._sessionInfo).to.deep.equal({
+            accessInfo: expectedResponse
+          });
+        });
+      });
+
+      it('returns a fulfilled promise with a success message', function() {
+        return promise.then((response) => {
+          expect(response).to.equal('Contxt Authentication successful.');
```
greg5green: If you don't change `_getApiToken` to return the token, you should make this `'Contxt Authentication successful.'` a generated variable so it is more clear exactly what is going on

src/sessionTypes/cliAuth.js
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Saves session info under a specific key for future use
+   *
+   * @param {string} key
+   * @param {Object} sessionInfo
+   *
+   * @private
+   */
+  _saveSession(key, sessionInfo) {
+    this._sessionInfo[key] = sessionInfo;
```
greg5green: What's the benefit of saving it under a certain key?

src/sessionTypes/cliAuth.js
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @typicalname contxtSdk.auth
+ *
+ * @example
+ * const ContxtSdk = require('@ndustrial/contxt-sdk');
+ *
+ * const contxtService = new ContxtSdk({
+ *   config: {
+ *     auth: {
+ *       clientId: 'bleED0RUwb7CJ9j7D48tqSiSZRZn29AV'
```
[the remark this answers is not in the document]
leeallen337: Absolutely 👍

src/sessionTypes/cliAuth.js
leeallen337: I'm fine with resolving with the access token. I think I initially had this resolving it with a string to be logged out by the `contxt-cli` before adding in the `_getApiToken` in the `.then` and just forgot. 👍

src/sessionTypes/cliAuth.js
leeallen337: I was just thinking of being consistent between all the public API methods on this module. If some of them are going to be promise-based then why not support it for all public facing methods? Maybe a different developer who is using this wants it to be `then`-able and handle something else after logging out
```
return contxtService.auth
  .logout()
  .then(() => {
    // some action someone else wants to do here
  })
```
What do you think?

src/sessionTypes/cliAuth.js
leeallen337: Discussed in-person

src/sessionTypes/cliAuth.spec.js
leeallen337: Updated to return the token 👍

src/sessionTypes/cliAuth.js
greg5green: I don't hate it, just it's only for this session type. The web auth one isn't even async for `logIn`

src/sessionTypes/passwordGrantAuth.js
```diff
@@ -0,0 +1,182 @@
...
+        const { data } = response;
+
+        this._saveSession({
+          accessToken: data.access_token
+        });
+
+        return data.access_token;
+      })
+      .catch(() => {
+        throw new Error('An error occurred during authorization with Contxt.');
```
greg5green: What do you think about getting rid of this catch and letting the error just propagate? The WebAuth one does that and it's nice. The Machine Auth one does not, but uses a different error message: `'There was a problem getting a token from the ContxtAuth server. Please check your configuration settings.'`. If not getting rid of the catch, could you change the text?

leeallen337:
@greg5green I had one more commit for running the doc build. Do you have anymore comments? If not, could you approve?

src/sessionTypes/passwordGrantAuth.spec.js
```diff
@@ -0,0 +1,462 @@
...
+        let expectedApiToken;
+        let post;
+        let promise;
+        let saveSession;
+
+        beforeEach(function() {
+          accessToken = faker.internet.password();
+          expectedApiToken = faker.internet.password();
+
+          post = this.sandbox.stub(axios, 'post').callsFake(() => {
```
greg5green: Why not `post = this.sandbox.stub(axios, 'post').resolves({ data: { access_token: expectedApiToken }  });`?

greg5green: (this is very minor, im sorry)

leeallen337: Good call. I think I had done it this way first and then started using `resolve`. I'll update this

src/sessionTypes/passwordGrantAuth.js
leeallen337: I like the idea of just letting it propagate. I'll drop the `catch` here similar to the WebAuth.

src/sessionTypes/passwordGrantAuth.js
```diff
@@ -0,0 +1,182 @@
...
+   * Gets the current API token (used to communicate with other Contxt APIs)
+   *
+   * @returns {Promise}
+   * @fulfills {string} apiToken
+   * @rejects {Error}
+   */
+  getCurrentApiToken() {
+    return new Promise((resolve, reject) => {
+      if (!this._sessionInfo.accessToken) {
+        return reject(new Error('No access token found.'));
```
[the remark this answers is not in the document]
leeallen337: Updated 👍

PR #86, opened 2019-02-12
Message Bus authorization for publishing/subscribing to a channel

erichardson30:
## Why?
As a developer I need to send an authorization message to the message bus for publishing/subscribing to a channel [MOB-2339](https://lineagelogistics.atlassian.net/browse/MOB-2339)

## What changed?
Added in a function to the WebSocketConnection to authorize a channel
Sends message to message bus via open WebSocket connection to authorize channel
Testing added


src/bus/webSocketConnection.js
```diff
@@ -9,10 +9,89 @@ class WebSocketConnection {
...
+   *         console.log(res.authorized)
+   *       })
+   *       .catch((authErrorEvent) => {
+   *         console.log(authErrorEvent)
+   *       });
+   *     })
+   *     .catch((errorEvent) => {
+   *       console.log(errorEvent);
+   *     });
+   * });
```
greg5green: I think you can un-nest these `then`s a bit:


```
axios
  .post(
    ["https://", SERVICE_ADDRESS, "/authorizations/channels"].join(""),
    {
      resources: [
        {
          resource: "CHANNEL_YOU_NEED",
          actions: ["publish", "subscribe"]
        }
      ]
    },
    {
      headers: {
        Authorization: ["Bearer", ACCESS_TOKEN_FROM_PREVIOUS_STEP].join(" ")
      }
    }
  )
  .then(response => {
    const token = response.data;

    return contxtSdk.bus
      .connect("4f0e51c6-728b-4892-9863-6d002e61204d")
      .then((webSocket) => webSocket.authorize(token));
  })
  .then(response) => {
    if (response.error) {
      console.log("Error: ", response.error);
    }

    console.log(response.authorized);
  }
  .catch((error) => console.log("Error: ", error));
```

Also, have you thought about not including the info about how to go and "get" the token? It should be documented, but probably at a higher level -- it doesn't really have much to do with this method (outside of consuming the response, of course)

src/bus/webSocketConnection.js
```diff
@@ -9,10 +9,89 @@ class WebSocketConnection {
...
+      if (!this._webSocket || !this._webSocket.OPEN) {
+        return reject(new Error('WebSocket connection not open'));
+      }
+
+      this._webSocket.onmessage = (message) => {
+        const messageData = JSON.parse(message.data);
+        const error = messageData.error || null;
+        const authorized = !error;
+
+        return resolve({ authorized, error });
```
greg5green: Can you make this `isAuthorized`?

greg5green: I think you might need to tear down these onmessage/onerror handlers too

src/bus/webSocketConnection.spec.js
```diff
@@ -36,6 +39,335 @@ describe('WebScoketConnection', function() {
...
+        let jsonRpcId;
+        let promise;
+        let send;
+        let token;
+        let ws;
+
+        beforeEach(function() {
+          expectedMessage = { jsonrpc: '2.0', id: '0', result: null };
+          expectedOrganization = fixture.build('organization');
+          send = this.sandbox.stub(expectedWebSocket, 'send');
```
greg5green: I don't think you need to stub this -- we can utilize the features of mock-socket more. I don't think mock-socket does as much as I think sometimes, but could just spy here

src/bus/webSocketConnection.spec.js
```diff
@@ -36,6 +39,335 @@ describe('WebScoketConnection', function() {
...
+          token = faker.random.uuid();
```
greg5green: We've generally been using `faker.internet.password()` for tokens since it looks more like the real thing

src/bus/webSocketConnection.spec.js
```diff
@@ -36,6 +39,335 @@ describe('WebScoketConnection', function() {
...
+        let expectedOrganization;
+        let expectedJsonRpc;
+        let jsonRpcId;
+        let promise;
+        let send;
+        let token;
+        let ws;
+
+        beforeEach(function() {
+          expectedMessage = { jsonrpc: '2.0', id: '0', result: null };
```
greg5green: Shouldn't the `id` be the same as the `_jsonRpcId`

src/bus/webSocketConnection.spec.js
```diff
@@ -36,6 +39,335 @@ describe('WebScoketConnection', function() {
...
+          return promise.then(() => {
+            expect(ws._jsonRpcId).to.equal(jsonRpcId + 1);
+          });
+        });
+
+        it('fulfills the promise', function() {
+          expect(promise).to.be.fulfilled;
+        });
+
+        it('recieves a message from the server', function() {
```
greg5green: I'd only test the differences in this failure state, not the stuff that happens either way

src/bus/webSocketConnection.js
greg5green: If there's an error, would it be more appropriate to reject?

erichardson30: I went back and forth with the error resolve/reject for a while. It was kinda weird because if you're unauthorized it comes back in the "error" even though there isn't technically an error

greg5green: Since this method abstracts away the message bus portion of things, it's weird to me that it wouldn't reject. You wouldn't expect your promise's `then` to get an error

erichardson30: Yeah that makes sense. Will update

PR #87, opened 2019-02-14
Publish a message to the message bus on a specific channel

erichardson30:
## Why?
As a developer I need to be able to publish a message to the message bus on a particular channel. 
[MOB-1629](https://lineagelogistics.atlassian.net/browse/MOB-1629)

## What changed?
* Added a publish function to publish to the message bus
* Added tests

### DIFF
--- a/src/bus/webSocketConnection.js
+++ b/src/bus/webSocketConnection.js
@@ -112,7 +112,7 @@ class WebSocketConnection {
    *
    * @param {string} serviceClientId Client ID of the message bus service
    * @param {string} channel Message bus channel the message is being sent to
-   * @param {Object} message Message being sent to the message bus
+   * @param {Any} message Message being sent to the message bus. Must be valid JSON.
    *
    * @returns {Promise}
    * @fulfill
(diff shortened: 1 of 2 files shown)


src/bus/webSocketConnection.js
```diff
@@ -106,6 +106,97 @@ class WebSocketConnection {
   close() {
     this._webSocket.close();
   }
+
+  /**
+   * Publishes a message to a specific channel on the message bus
+   *
+   * @param {string} serviceClientId Client ID of the message bus service
+   * @param {string} channel Message bus channel the message is being sent to
+   * @param {Object} message Message being sent to the message bus
```
greg5green: Could this actually be Any? I'm thinking it doesn't matter what you send in as long as it is JSON encodeable

src/bus/webSocketConnection.spec.js
```diff
@@ -442,4 +441,527 @@ describe('Bus/WebSocketConnection', function() {
...
+      });
+
+      it('rejects the promise', function() {
+        return expect(promise).to.be.rejectedWith(
+          'A channel is required for publishing'
+        );
+      });
+    });
+
+    context('when there is not a message sent', function() {
```
greg5green: Is there an easy way to do some code reuse on these tests that are almost all the same?

src/bus/webSocketConnection.js
erichardson30: Yes, you're correct

PR #88, opened 2019-02-18
Get a WebSocket connection for an organization

erichardson30:
## Why?
As a user I need to get a WebSocket connection for an organization

## What changed?
Gets a WebSocket connection for an organization
Added tests

### DIFF
--- a/src/bus/index.js
+++ b/src/bus/index.js
@@ -118,6 +118,21 @@ class Bus {
         });
     });
   }
+
+  /**
+   * Gets the WebSocketConnection for an organization id
+   * If a connection already exists for that organization id, the connection is returned, otherwise returns undefined.
+   *
+   * @param {string} organizationId UUID corresponding with an organization
+   *
+   * @returns {WebSocketConnection}
+   *
+   * @example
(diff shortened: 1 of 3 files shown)


src/bus/index.js
```diff
@@ -118,6 +118,21 @@ class Bus {
         });
     });
   }
+
+  /**
+   * Gets the WebSocketConnection for an organization id
+   * If a connection already exists for that organization id, the connection is returned, otherwise returns undefined.
```
greg5green: This is a good way to put the behavior

src/bus/index.spec.js
```diff
@@ -298,4 +298,67 @@ describe('Bus', function() {
...
+        let expectedReturnedSocket;
+
+        beforeEach(function() {
+          expectedReturnedSocket = bus.getWebSocketConnection(
+            expectedOrganization.id
+          );
+        });
+
+        it('returns the socket', function() {
+          expect(expectedReturnedSocket).to.deep.equal(
```
greg5green: What do you think about changing the way this is named? It seems like `bus._webSockets[expectedOrganization.id]` should be the expected socket and the return should just be the web socket connection? Like, the return isn't what is expected and now I'm rambling

erichardson30: 👍

PR #90, opened 2019-02-25
Refactor onmessage and onerror handling for message bus web socket

erichardson30:
## Why?
The onmessage and onerror handlers were not robust enough to handle multiple messages coming in at the same time. 

## What changed?
* Each message that the SDK sends to the message bus gets its own onmessage handler. 
* The handler is removed after the message is processed.
* Tests updated.

### DIFF
--- a/src/bus/webSocketConnection.js
+++ b/src/bus/webSocketConnection.js
@@ -1,3 +1,5 @@
+import uuid from 'uuid/v4';
+
 /**
  * Module that wraps the websocket connection to the message bus
  * to provide the developer with a specific set of functionality
@@ -9,9 +11,15 @@ class WebSocketConnection {
    * @param {string} organizationId UUID corresponding with an organization
    */
   constructor(webSocket, organizationId) {
-    this._jsonRpcId = 0;
+    this._jsonRpcId = null;
+    this._messageHandlers = {};
     this._organizationId = organizationId;
     this._webSocket = webSocket;
+
+    if (this._webSocket) {
+      this._webSocket.onmessage = this.onMessage;
+      this._webSocket.onerror = this.onError;
+    }
   }
 
   /**
@@ -41,8 +49,6 @@ class WebSocketConnection {
         return reject(new Error('A token is required for authorization'));
       }
 
-      const messageId = this._jsonRpcId;
-
       if (
         !this._webSocket ||
         this._webSocket.readyState !== this._webSocket.OPEN
@@ -50,30 +56,19 @@ class WebSocketConnection {
         return reject(new Error('WebSocket connection not open'));
       }
 
-      this._webSocket.onmessage = (message) => {
-        const messageData = JSON.parse(message.data);
-        const error = messageData.error;
+      this._jsonRpcId = uuid();
+      const messageId = this._jsonRpcId;
+
+      this._messageHandlers[messageId] = (message) => {
+        const error = message.error;
+
+        delete this._messageHandlers[messageId];
 
         if (error) {
-          this._webSocket.onmessage = null;
-          this._webSocket.onerror = null;
-
           return reject(error);
         }
 
-        if (parseInt(messageData.id) === messageId) {
-          this._webSocket.onmessage = null;
-          this._webSocket.onerror = null;
-
-          return resolve();
-        }
-      };
-
-      this._webSocket.onerror = (errorEvent) => {
-        this._webSocket.onmessage = null;
-        this._webSocket.onerror = null;
-
-        return reject(errorEvent);
+        return resolve();
       };
 
       this._webSocket.send(
@@ -83,11 +78,9 @@ class WebSocketConnection {
           params: {
             token
           },
-          id: this._jsonRpcId
+          id: messageId
         })
       );
-
-      this._jsonRpcId++;
     });
   }
 
@@ -106,6 +99,37 @@ class WebSocketConnection {
   close() {
     this._webSocket.close();
   }
+
+  /**
+   * Handles WebSocket errors.
+   * The `ws` library also closes the socket when an error occurs.
+   * Since the socket connection closes, the jsonRpcId and message handlers are cleared
+   *
+   */
+  onError = (error) => {
+    this._jsonRpcId = null;
+    this._messageHandlers = {};
+
+    throw new Error(error);
+  };
+
+  /**
+   * Handles messages sent from the Message Bus WebSocket connection.
+   *
+   */
+  onMessage = (message) => {
+    let messageData;
+
+    try {
+      messageData = JSON.parse(message.data);
+    } catch (err) {
+      throw new Error('Invalid JSON in message');
+    }
+
+    if (this._messageHandlers[messageData.id]) {
+      this._messageHandlers[messageData.id](messageData);
+    }
+  };
 
   /**
    * Publishes a message to a specific channel on the message bus
@@ -146,8 +170,6 @@ class WebSocketConnection {
         return reject(new Error('A message is required for publishing'));
       }
 
-      const messageId = this._jsonRpcId;
-
       if (
         !this._webSocket ||
         this._webSocket.readyState !== this._webSocket.OPEN
@@ -155,30 +177,18 @@ class WebSocketConnection {
         return reject(new Error('WebSocket connection not open'));
       }
 
-      this._webSocket.onmessage = (message) => {
-        const messageData = JSON.parse(message.data);
-        const error = messageData.error;
+      this._jsonRpcId = uuid();
+      const messageId = this._jsonRpcId;
+
+      this._messageHandlers[messageId] = (message) => {
+        const error = message.error;
+        delete this._messageHandlers[messageId];
 
         if (error) {
-          this._webSocket.onmessage = null;
-          this._webSocket.onerror = null;
-
           return reject(error);
         }
 
-        if (parseInt(messageData.id) === messageId) {
-          this._webSocket.onmessage = null;
-          this._webSocket.onerror = null;
-
-          return resolve();
-        }
-      };
-
-      this._webSocket.onerror = (errorEvent) => {
-        this._webSocket.onmessage = null;
-        this._webSocket.onerror = null;
-
-        return reject(errorEvent);
+        return resolve();
       };
 
       this._webSocket.send(
@@ -190,11 +200,9 @@ class WebSocketConnection {
             channel,
             message
           },
-          id: this._jsonRpcId
+          id: messageId
         })
       );
-
-      this._jsonRpcId++;
     });
   }
 }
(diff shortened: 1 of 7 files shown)


src/bus/webSocketConnection.spec.js
```diff
@@ -442,6 +325,176 @@ describe('Bus/WebSocketConnection', function() {
...
+            JSON.stringify({
+              jsonrpc: '2.0',
+              id: expectedUUID,
+              result: null
+            })
+          );
+        });
+
+        it('calls the onmessage function for the message handler', function() {
+          expect(expectedMessageHandlers[expectedUUID]).to.be.calledOnce;
```
greg5green: What do you think about checking that the message is passed on too?

src/bus/webSocketConnection.spec.js
```diff
@@ -442,6 +325,176 @@ describe('Bus/WebSocketConnection', function() {
...
+        };
+
+        onError = this.sandbox.spy(ws, 'onError');
+      });
+
+      it('throws an error', function() {
+        try {
+          webSocketServer.emit('error');
+        } catch (err) {
+          expect(onError).to.throw(Error);
```
greg5green: Why would the ws `onError` throw like that? And why would we be testing it?

src/bus/webSocketConnection.spec.js
```diff
@@ -442,6 +325,176 @@ describe('Bus/WebSocketConnection', function() {
...
+        ws._messageHandlers = {
+          [faker.random.uuid()]: this.sandbox.stub()
+        };
+
+        onError = this.sandbox.spy(ws, 'onError');
+      });
+
+      it('throws an error', function() {
+        try {
+          webSocketServer.emit('error');
```
greg5green: What do you think about just invoking this?

src/bus/webSocketConnection.spec.js
```diff
@@ -442,6 +325,176 @@ describe('Bus/WebSocketConnection', function() {
...
+          };
+
+          ws = new WebSocketConnection(
+            expectedWebSocket,
+            expectedOrganization.id
+          );
+
+          ws._messageHandlers = expectedMessageHandlers;
+
+          webSocketServer.emit(
```
greg5green: What do you think about just invoking `onMessage`? The other, previous cases (that also have comments about this same thing) were more of a call + response test because we had to have an open connection for it to work

src/bus/webSocketConnection.spec.js
```diff
@@ -74,11 +76,13 @@ describe('Bus/WebSocketConnection', function() {
             id: jsonRpcId
           });
 
-          promise = ws.authorize(token);
-
           webSocketServer.emit(
             'message',
-            JSON.stringify({ jsonrpc: '2.0', id: jsonRpcId, result: null })
+            JSON.stringify({
```
greg5green: To test this unit, we don't really need to do this anymore -- can test it by invoking the method directly. The mechanism for how onmessage works isn't even in this method anymore

src/bus/webSocketConnection.spec.js
```diff
@@ -442,6 +325,176 @@ describe('Bus/WebSocketConnection', function() {
...
+
+        it('does not call the onmessage function for the message handler', function() {
+          try {
+            webSocketServer.emit('message', {
+              jsonrpc: '2.0',
+              id: expectedUUID,
+              result: null
+            });
+          } catch (err) {
+            expect(expectedMessageHandlers[expectedUUID]).to.not.be.called;
```
greg5green: Is this really being executed?

src/bus/webSocketConnection.spec.js
```diff
@@ -496,27 +549,15 @@ describe('Bus/WebSocketConnection', function() {
...
-        it('tears down the onmessage handler', function() {
-          return promise.then(function() {
-            expect(ws._webSocket.onmessage).to.be.undefined;
-          });
+        it('fulfills the promise', function() {
+          expect(promise).to.be.fulfilled;
         });
 
-        it('tears down the onerror handler', function() {
+        it('tears down the on message and on error handlers', function() {
```
greg5green: This description is also not accurate

src/bus/webSocketConnection.spec.js
```diff
@@ -809,17 +759,17 @@ describe('Bus/WebSocketConnection', function() {
...
-        });
-      });
-
       it('rejects the promise', function() {
         return expect(promise).to.be.rejectedWith(
           'WebSocket connection not open'
         );
       });
+
+      it('does not create the on message or on error handlers', function() {
```
greg5green: Another one

src/bus/webSocketConnection.spec.js
```diff
@@ -857,17 +807,17 @@ describe('Bus/WebSocketConnection', function() {
...
-        });
-      });
-
       it('rejects the promise', function() {
         return expect(promise).to.be.rejectedWith(
           'A service client id is required for publishing'
         );
       });
+
+      it('does not create the on message or on error handlers', function() {
```
greg5green: Here too

src/bus/webSocketConnection.spec.js
```diff
@@ -905,17 +855,17 @@ describe('Bus/WebSocketConnection', function() {
...
-        });
-      });
-
       it('rejects the promise', function() {
         return expect(promise).to.be.rejectedWith(
           'A channel is required for publishing'
         );
       });
+
+      it('does not create the on message or on error handlers', function() {
```
greg5green: omg how many times does this even happen

src/bus/webSocketConnection.spec.js
```diff
@@ -951,17 +901,17 @@ describe('Bus/WebSocketConnection', function() {
...
-        });
-      });
-
       it('rejects the promise', function() {
         return expect(promise).to.be.rejectedWith(
           'A message is required for publishing'
         );
       });
+
+      it('does not create the on message or on error handlers', function() {
```
greg5green: i swear this must just be copy and pasted a bunch of times

src/bus/webSocketConnection.spec.js
```diff
@@ -4,7 +4,7 @@ import WebSocketConnection from './webSocketConnection';
 
 const DELAY = 5;
 
-describe('Bus/WebSocketConnection', function() {
+describe.only('Bus/WebSocketConnection', function() {
```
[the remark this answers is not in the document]
erichardson30: UGH! I always forget to remove this 🤦‍♂️

src/bus/webSocketConnection.js
```diff
@@ -107,6 +100,37 @@ class WebSocketConnection {
...
+  /**
+   * Handles WebSocket errors.
+   * The `ws` library also closes the socket when an error occurs.
+   * Since the socket connection closes, the jsonRpcId and message handlers are cleared
+   *
+   */
+  onError = (error) => {
+    this._jsonRpcId = null;
+    this._messageHandlers = {};
+
```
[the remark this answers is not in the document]
erichardson30: I don't think so. The websocket connection will be closed so the user won't be able to do anything else without establishing a new connection...but down the road if we were to implement some reconnection stuff, we probably want access to the socket without setting it to `null`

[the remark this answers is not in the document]
greg5green: 👍

src/bus/webSocketConnection.js
```diff
@@ -107,6 +98,34 @@ class WebSocketConnection {
...
 
+  /**
+   * Handles WebSocket errors.
+   * The `ws` library also closes the socket when an error occurs.
+   * Since the socket connection closes, the jsonRpcId and message handlers are cleared
+   */
+  onError = (error) => {
+    this._messageHandlers = {};
+
+    console.log(error);
```
greg5green: It would be good to provide some context here, like `Error from this bus thing: ${error}`

src/bus/webSocketConnection.js
```diff
@@ -107,6 +98,34 @@ class WebSocketConnection {
     this._webSocket.close();
   }
 
+  /**
+   * Handles WebSocket errors.
+   * The `ws` library also closes the socket when an error occurs.
+   * Since the socket connection closes, the jsonRpcId and message handlers are cleared
+   */
+  onError = (error) => {
```
greg5green: Need to add info on the `error` parameter. And what do you think about making this a "private" method? Users shouldn't really be calling it

src/bus/webSocketConnection.js
```diff
@@ -107,6 +98,34 @@ class WebSocketConnection {
...
+    this._messageHandlers = {};
+
+    console.log(error);
+  };
+
+  /**
+   * Handles messages sent from the Message Bus WebSocket connection.
+   */
+  onMessage = (message) => {
```
greg5green: Need to add info on the `message` parameter. And what do you think about making this a "private" method? Users shouldn't really be calling it

src/bus/webSocketConnection.spec.js
greg5green: This was marked resolved but there's no comment about it and nothing has changed?

src/bus/webSocketConnection.spec.js
```diff
@@ -442,6 +325,139 @@ describe('Bus/WebSocketConnection', function() {
...
+      ws._messageHandlers = {
+        [faker.random.uuid()]: this.sandbox.stub()
+      };
+    });
+
+    it('resets the messageHandlers', function() {
+      try {
+        ws.onError(expectedError);
+      } catch (err) {
+        expect(ws._messageHandlers).to.be.empty;
```
greg5green: This expect never runs -- you aren't throwing anything!

src/bus/webSocketConnection.spec.js
```diff
@@ -442,6 +325,139 @@ describe('Bus/WebSocketConnection', function() {
...
+          onMessage = this.sandbox.spy(ws, 'onMessage');
+        });
+
+        it('throws an error', function() {
+          try {
+            ws.onMessage({ data: expectedMessage });
+          } catch (err) {
+            expect(onMessage).to.throw('Invalid JSON in message');
+          }
+        });
```
greg5green: Could you change this to:
```
it('throws an error', function() {
  expect(() => ws.onMessage({ data: expectedMessage })).to.throw(
    'Invalid JSON in message'
  );
});
```

Reasoning behind this? If you were to change your catch to do, say an early return, what is here will not get tested anymore, but the revised version would still show up as a failing test.

src/bus/webSocketConnection.spec.js
greg5green: There was no comment before, but you have the same situation in the publish -- it's no longer testing just a singular unit since it depends on `onMessage` to do the handling. Should be testing that that message handler is set up correctly and what the message handler itself does

greg5green:
You'll need to build the docs too.

src/bus/webSocketConnection.spec.js
erichardson30: I think I just misunderstood what you meant the first time. Should be updated now

PR #94, opened 2019-03-12
Create an event type

erichardson30:
## Why?
As a developer I need to create an event type.

## What changed?
* Added function in events module to create an event type
* Updated tests

### DIFF
--- a/src/events/index.js
+++ b/src/events/index.js
@@ -52,10 +52,11 @@ import { formatPaginatedDataFromServer } from '../utils/pagination';
  * @property {string} createdAt ISO 8601 Extended Format date/time string
  * @property {string} description
  * @property {string} id UUID
- * @property {boolean} isRealtimeEnabled
+ * @property {boolean} isOngoingEvent Flag for if the event is ongoing/updated frequently
+ * @property {boolean} isRealtimeEnabled Flag for if the event is real time
  * @property {number} level Priority level associated with event type
  * @property {string} name
- * @property {string} slug
+ * @property {string} slug Unique per clientId to identify the event type
  * @property {string} updatedAt ISO 8601 Extended Format date/time string
  */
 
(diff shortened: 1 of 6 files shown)


src/events/index.js
```diff
@@ -316,6 +317,64 @@ class Events {
...
+   * Method: POST
+   *
+   * @param {Object} eventType
+   * @param {string} eventType.name
+   * @param {string} eventType.description
+   * @param {integer} [eventType.level]
+   * @param {string} eventType.clientId
+   * @param {string} eventType.slug
+   * @param {boolean} eventType.isRealtimeEnabled
+   * @param {boolean} eventType.isOngoingEvent
```
greg5green: Can you add a little more info about these keys? There isn't a ton in the regular typedef either

src/events/index.js
```diff
@@ -316,6 +317,64 @@ class Events {
...
+  /**
+   * Creates a new event type
+   *
+   * API Endpoint: '/types'
+   * Method: POST
+   *
+   * @param {Object} eventType
+   * @param {string} eventType.name
+   * @param {string} eventType.description
+   * @param {integer} [eventType.level]
```
greg5green: We've been using `number` for the type since thats the only JS type

src/events/index.spec.js
```diff
@@ -525,4 +525,99 @@ describe('Events', function() {
...
+      it('creates a new event type', function() {
+        expect(request.post).to.be.deep.calledWith(
+          `${expectedHost}/types`,
+          eventTypeToServerAfterFormat
+        );
+      });
+
+      it('formats the returned object', function() {
+        return promise.then(() => {
+          expect(toCamelCase).to.be.deep.calledWith(
```
greg5green: Why would all these tests be `deep`?

erichardson30: Keeping consistent with all the other `create` tests.

erichardson30: Updated

PR #97, opened 2019-03-25
Feature/favorite-applications

smstromb:
## Why?
User needs to create, delete, and read their favorited applications
https://lineagelogistics.atlassian.net/browse/MOB-2443

## What changed?
Added:
- Coordinator.createFavoriteApplication()
- Coordinator.deleteFavoriteApplication()
- Coordinator.getFavoriteApplications()
- ContxtUserFavoriteApplication fixture

### DIFF
--- a/src/coordinator/index.js
+++ b/src/coordinator/index.js
@@ -40,7 +40,7 @@ import { toCamelCase } from '../utils/objects';
 
 /**
  * @typedef {Object} ContxtUserFavoriteApplication
- * @property {string} applicationId
+ * @property {number} applicationId
  * @property {string} createdAt ISO 8601 Extended Format date/time string
  * @property {string} id
  * @property {string} updatedAt ISO 8601 Extended Format date/time string
@@ -68,10 +68,12 @@ class Coordinator {
   }
 
   /**
-   * Creates an application to a user's favorited applications
+   * Adds an application to the current user's list of favorited applications
    *
    * API Endpoint: '/applications/:applicationId/favorites'
    * Method: POST
+   *
+   * Note: Only valid for web user's using auth0WebAuth session type
    *
    * @param {number} applicationId The ID of the application
    *
@@ -100,10 +102,12 @@ class Coordinator {
   }
 
   /**
-   * Deletes an application from a user's favorited applications
+   * Removes an application from the current user's list of favorited applications
    *
    * API Endpoint: '/applications/:applicationId/favorites'
    * Method: DELETE
+   *
+   * Note: Only valid for web user's using auth0WebAuth session type
    *
    * @param {number} applicationId The ID of the application
    *
@@ -175,11 +179,12 @@ class Coordinator {
   }
 
   /**
-   * Gets a user's list of favorited applications
+   * Gets the current user's list of favorited applications
    *
    * API Endpoint: '/applications/favorites'
    * Method: GET
    *
+   * Note: Only valid for web user's using auth0WebAuth session type
    *
    * @returns {Promise}
    * @fulfill {ContxtUserFavoriteApplication[]} A list of favorited applications
@@ -194,9 +199,7 @@ class Coordinator {
   getFavoriteApplications() {
     return this._request
       .get(`${this._baseUrl}/applications/favorites`)
-      .then((favoriteApps) =>
-        favoriteApps.map((favoriteApp) => toCamelCase(favoriteApp))
-      );
+      .then((favoriteApps) => toCamelCase(favoriteApps));
   }
 
   /**
--- a/src/coordinator/index.js
+++ b/src/coordinator/index.js
@@ -76,7 +76,7 @@ class Coordinator {
    * @param {number} applicationId The ID of the application
    *
    * @returns {Promise}
-   * @fulfill {ContxtUserFavoriteApplication[]} Information about the contxt application favorite
+   * @fulfill {ContxtUserFavoriteApplication} Information about the contxt application favorite
    * @reject {Error}
    *
    * @example
--- a/src/coordinator/index.js
+++ b/src/coordinator/index.js
@@ -81,11 +81,11 @@ class Coordinator {
    *
    * @example
    * contxtSdk.coordinator
-   *   .createApplicationFavorite(25)
-   *   .then((applicationFavorite) => console.log(applicationFavorite))
+   *   .createFavoriteApplication(25)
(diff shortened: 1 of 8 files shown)


src/coordinator/index.js
```diff
@@ -58,6 +67,69 @@ class Coordinator {
     this.edgeNodes = new EdgeNodes(sdk, request, baseUrl);
   }
 
+  /**
+   * Creates an application to a user's favorited applications
```
greg5green: Can you add some comment about how this only works for Web users using the `auth0WebAuth` session type? (For all of these methods)

It should also probably be noted that these are for the _current_ user, not _a_ user

src/coordinator/index.js
```diff
@@ -102,6 +174,31 @@ class Coordinator {
...
+   * contxtSdk.coordinator
+   *   .getFavoriteApplications()
+   *   .then((favoriteApplications) => console.log(favoriteApplications))
+   *   .catch((err) => console.log(err));
+   */
+  getFavoriteApplications() {
+    return this._request
+      .get(`${this._baseUrl}/applications/favorites`)
+      .then((favoriteApps) =>
+        favoriteApps.map((favoriteApp) => toCamelCase(favoriteApp))
```
greg5green: You can actually just feed `favoriteApps` into `toCamelCase`! (https://github.com/ndustrialio/contxt-sdk-js/blob/master/src/utils/objects/createCaseChangeFn.js#L59)

src/coordinator/index.js
leeallen337: Perhaps I'm reading this incorrectly but this wording seems weird to me. Is this actually creating an application or is this just marking an application as a favorite of a specific user?

What about something like `Adds an application to a user's list of favorited applications`?

src/coordinator/index.spec.js
```diff
@@ -51,6 +51,111 @@ describe('Coordinator', function() {
...
+          }
+        );
+
+        request = {
+          ...baseRequest,
+          post: this.sandbox.stub().resolves(applicationFavoriteFromServer)
+        };
+        toCamelCase = this.sandbox
+          .stub(objectUtils, 'toCamelCase')
+          .callsFake((app) => expectedApplicationFavorite);
```
greg5green: This works just fine so it doesn't have to change, but could use `returns` here instead of `callsFake`

src/coordinator/index.js
```diff
@@ -58,6 +67,69 @@ class Coordinator {
...
+      );
+    }
+
+    return this._request
+      .post(`${this._baseUrl}/applications/${applicationId}/favorites`)
+      .then((favoriteApplication) => toCamelCase(favoriteApplication));
+  }
+
+  /**
+   * Deletes an application from a user's favorited applications
```
leeallen337: What about `Removes an application from a user's list of favorited applications`?

support/fixtures/factories/contxtUserFavoriteApplication.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const factory = require('rosie').Factory;
+const faker = require('faker');
+
+factory
+  .define('contxtUserFavoriteApplication')
+  .sequence('id')
```
greg5green: This should be a UUID

src/coordinator/index.js
```diff
@@ -24,18 +39,12 @@ import { toCamelCase } from '../utils/objects';
  */
 
 /**
- * @typedef {Object} ContxtApplication
- * @property {string} clientId
- * @property {string} clientSecret
+ * @typedef {Object} ContxtUserFavoriteApplication
+ * @property {string} applicationId
```
greg5green: I think this is a number

src/coordinator/index.js
smstromb: Thanks, I good real English I speak...

By all these methods, do you mean all of contxt coordinator or just the favorite application routes?

src/coordinator/index.spec.js
```diff
@@ -51,6 +51,111 @@ describe('Coordinator', function() {
...
+
+        const coordinator = new Coordinator(baseSdk, request);
+        coordinator._baseUrl = expectedHost;
+        promise = coordinator.createFavoriteApplication(
+          expectedApplicationFavorite.applicationId
```
leeallen337: This is fine but it feels a little weird since the `createFavoriteApplication` is to be called with the application id and not using the application favorite application id. Especially since you're doing it with the regular application in the DELETE [here](https://github.com/ndustrialio/contxt-sdk-js/pull/97/files#diff-b4dd03a165bf28f14cfa732bcf0c235eR128)

Maybe before line 63 you can do the 
```
application = fixture.build('contxtApplication');
```
and then have
```
expectedApplicationFavorite = fixture.build(
  'contxtUserFavoriteApplication',
  {  applicationId: application.id }
);
```
or something

leeallen337: Maybe I'm just being weird. You don't have to make this change if you don't want to

smstromb: I had the exact same thought as I was using the application fixture for the DELETE test but forgot to go back and change it. 

I'll fix, thanks

src/coordinator/index.js
greg5green: I think just the favorite application routes

smstromb: I've added a comment warning about the session type requirements to those 3 functions. Will you also double check that the same isn't true about getAllOrganizations() and getAllApplications()?

src/coordinator/index.js
smstromb: Done 👍

src/coordinator/index.spec.js
smstromb: Fixed 👍

support/fixtures/factories/contxtUserFavoriteApplication.js
smstromb: Fixed 👍

src/coordinator/index.js
smstromb: You are correct, changed to number

src/coordinator/index.js
smstromb: 👍

src/coordinator/index.js
smstromb: 👍

src/coordinator/index.spec.js
smstromb: Fixed 👍

PR #103, opened 2019-05-07
Switch to On-Demand Access Tokens for auth0WebAuth session type

greg5green:
## Why?
Currently, when authenticating through Auth0, we get attempt to get one token that has permissions and scopes for every API that the SDK can interact with. With new needs, we need something that is a little more configurable and two options seemed good:

1. Getting many tokens [on demand] that are each specific to an API. When attempting to make a request to an API, it will check if there is already a token and use that or request a token if there isn't one. If a particular project doesn't use an API, it never asks for a token for that API 
2. Keep only one access token for access to contxt APIs, but have some mechanism where a developer can define when APIs they are going to interact with

Option one has the advantage of working very, very similarly to how the machine to machine authentication works (and is probably easier to implement) so we have gone down that route.

## What changed?
- Implemented on-demand, individual API tokens for the `auth0WebAuth` session type (option one)


src/sessionTypes/auth0WebAuth.js
```diff
@@ -137,28 +156,28 @@ class Auth0WebAuth {
...
-      });
+        }
+      );
     });
   }
 
   /**
-   * Routine that takes unparsed information from Auth0, uses it to get a valid API token, and then
+   * Routine that takes unparsed information from Auth0, stores it in a way that
+   * can be used for getting access tokens, schedules it's future renewal, and
```
rob5standingby: `it's` -> `its`

greg5green: i'm so mad at myself

src/sessionTypes/auth0WebAuth.js
```diff
@@ -193,24 +209,24 @@ class Auth0WebAuth {
...
 
     localStorage.removeItem('access_token');
-    localStorage.removeItem('api_token');
     localStorage.removeItem('expires_at');
 
-    clearTimeout(this._tokenRenewalTimeout);
+    clearTimeout(this._sessionRenewalTimeout);
 
-    this._onRedirect('/');
+    this._auth0.logout({ returnTo: new URL(window.location).origin });
```
rob5standingby: This isn't necessarily an issue but `URL` isn't supported in any version of IE. Given that the SDK could be used by some third party, do we want to try and support IE? Feel free to tell me to go away.

src/sessionTypes/auth0WebAuth.js
```diff
@@ -175,14 +194,11 @@ class Auth0WebAuth {
    * @returns {boolean}
    */
   isAuthenticated() {
-    const hasTokens = !!(
+    return !!(
       this._sessionInfo &&
       this._sessionInfo.accessToken &&
-      this._sessionInfo.apiToken &&
-      this._sessionInfo.expiresAt
+      this._sessionInfo.expiresAt > Date.now()
```
rob5standingby: 👍

src/sessionTypes/auth0WebAuth.js
greg5green: `import URL from 'url-parse';` is at the top of the file -- it's actually this package with IE support down to IE9: https://github.com/unshiftio/url-parse#readme

rob5standingby: Carry on sir

rob5standingby: Sorry, I see it now... It wasn't in the diff because it was already in the file 🤦‍♂

PR #106, opened 2019-05-17
MOB-2540: Feature/add-remove-roles

smstromb:
### Why?
Contxt needs the ability to add and remove roles from a user. [MOB-2540](https://lineagelogistics.atlassian.net/browse/MOB-2540)

### What changed?
- Added `Users#addRole` for adding a role to a user: `POST /users/:user_id/roles/:role_id` 
- Added `Users#removeRole` for removing a role from a user: `DELETE /users/:user_id/roles/:role_id`

### DIFF
--- a/src/coordinator/users.js
+++ b/src/coordinator/users.js
@@ -292,6 +292,42 @@ class Users {
       `${this._baseUrl}/organizations/${organizationId}/users/${userId}`
     );
   }
+
+  /**
+   * Removes a role from a user
+   *
+   * API Endpoint: '/users/:userId/roles/:roleId'
+   * Method: DELETE
+   *
+   * @param {string} userId The ID of the user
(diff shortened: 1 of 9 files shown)


src/coordinator/users.js
```diff
@@ -83,6 +93,43 @@ class Users {
...
+   * Adds a role to a user
+   *
+   * API Endpoint: '/users/:userId/roles/:roleId'
+   * Method: GET
+   *
+   * @param {string} userId The ID of the user
+   * @param {string} roleId The ID of the role
+   *
+   * @returns {Promise}
+   * @fulfill {ContxtUserRole[]} The newly created user role
```
greg5green: Should this just be a single role or an array of roles?

smstromb: Single, thanks!

PR #113, opened 2019-05-30
Adds Permissions#getOneByOrganizationId

smstromb:
## Why?
There's a new endpoint `GET /organizations/:organization_id/users/:user_id/permissions` that needs added to the SDK. 

## What changed?
- Added `Permissions#getOneByOrganizationId` (`GET /organizations/:organization_id/users/:user_id/permissions`) for getting a single user's permissions within an organization
- Renamed `Permissions#getByOrganizationId` to `Permissions#getAllByOrganizationId` (`GET /organizations/:organization_id/users/permissions`) for getting every user's permissions within an organization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## [v1.2.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.0) (2019-05-30)
+
+**Added**
+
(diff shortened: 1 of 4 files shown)


smstromb:
From IRL convo:

The renamed function is more or less only a week old and only used by contxt-react project so is safe to change.

PR #115, opened 2019-06-04
Updates user permissions fixture

smstromb:
## Why?

Updates contxtUserPermission fixture with most recent endpoint changes for https://github.com/ndustrialio/contxt-coordinator/pull/95

### DIFF
--- a/support/fixtures/factories/contxtUserPermissions.js
+++ b/support/fixtures/factories/contxtUserPermissions.js
@@ -2,6 +2,7 @@
 
 const factory = require('rosie').Factory;
 const faker = require('faker');
+const times = require('lodash.times');
 
 factory
   .define('contxtUserPermissions')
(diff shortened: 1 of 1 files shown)


support/fixtures/factories/contxtUserPermissions.js
```diff
@@ -45,10 +45,16 @@ factory
       permission.applications_implicit = permission.applicationsImplicit;
       delete permission.applicationsImplicit;
 
-      permission.stacks_explicit = permission.stacksExplicit;
+      permission.stacks_explicit = permission.stacksExplicit.map((s) => ({
```
greg5green: What about passing `fromServer` into `factory.buildList('contxtUserStack')` instead of having to map over stuff after the fact?

smstromb: I think it has to be mapped over after the fact. 

In tests, these server fixtures are built by overriding non-server fixtures:
```js
  const userPermissionFromServer = fixture.build(
    'contxtUserPermissions', 
     userPermission, 
    {
       fromServer: true
    });
```

which means the factory functions for the above fixture will never get called. They just assume the values of the `userPermission` fixture being passed in:

```js
  .attr('stacksExplicit', ['fromServer'], (fromServer) => {
    // Not called when overriding another fixture
    return factory
      .buildList(
        'contxtUserStack',
        faker.random.number({ min: 0, max: 15 }),
        null,
        { fromServer }
      )
      .map(
        (userStack) =>
          fromServer
            ? { id: userStack.stack_id, access_type: userStack.access_type }
            : { id: userStack.stackId, accessType: userStack.accessType }
      );
  })
```

It's also worth mentioning that these aren't actually `contxtUserStacks`, that fixture just conveniently had both `stack_id` and `access_type` types which seemed more efficient then building 2 different fixtures (`stack` and `userStack`) and then zipping them together. Either way, if I actually double down on using that fixture w/ `fromServer` then I'm going to have to map the `stack_id` -> `id` property somewhere.

```js
permission.stacks_explicit = permission.stacksExplicit.map(
  (stackAccess) => {
    const userStackFromServer = factory.build(
      'contxtUserStack',
      {
        stackId: stackAccess.id,
        accessType: stackAccess.accessType
      },
      {
        fromServer: true
      }
    );

    return {
      id: userStackFromServer.stack_id,
      access_type: userStackFromServer.access_type
    };
  }
);
delete permission.stacksExplicit;
```

greg5green: You're definitely right about the mapping over stuff! Hearing that it isn't actually a contxtUserStack makes me wonder if you should just be generate the values in this fixture factory -- my concern being future developer looking and saying "oh, this is a user stack, nbd" and all hell breaking loose.

PR #116, opened 2019-06-05
Feature/roles applications

melting:
### Why? 
As an admin of an organization, I want to be able to add and remove stacks and applications from my organization's roles. [MOB-2767](https://lineagelogistics.atlassian.net/browse/MOB-2767)

### What's Changed
- Added addStack and removeStack methods for roles
- Added addApplication and removeApplication methods for roles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## [v1.2.3](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.3) (2019-06-05)
+
+**Added**
+
+- Added `Coordinator.roles#addApplication` for adding an application to a role
+- Added `Coordinator.roles#addStack` for adding a stack to a role
+- Added `Coordinator.roles#removeApplication` for removing an application from a role
+- Added `Coordinator.roles#removeStack` for removing a stack from a role
+
 ## [v1.2.2](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.2) (2019-06-03)
 
 **Fixed**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [v1.2.2](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.2) (2019-06-03)
+
+**Fixed**
+
+- v1.2.1 was published without the changes introduced in 1.1.0. Republishing to re-add this code.
+
 ## [v1.2.1](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.1) (2019-05-30)
 
 **Added**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [v1.2.3](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.3) (2019-06-05)
+## [v1.2.4](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.4) (2019-06-06)
 
 **Added**
 
@@ -6,6 +6,12 @@
 - Added `Coordinator.roles#addStack` for adding a stack to a role
 - Added `Coordinator.roles#removeApplication` for removing an application from a role
 - Added `Coordinator.roles#removeStack` for removing a stack from a role
+
+## [v1.2.3](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.3) (2019-06-05)
+
+**Fixed**
+
+- Updated internal config builder to include `webSocket` key
 
 ## [v1.2.2](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.2) (2019-06-03)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [v1.2.4](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.4) (2019-06-06)
+## [v1.3.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.3.0) (2019-06-06)
 
 **Added**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [v1.3.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.3.0) (2019-06-06)
+## [v1.3.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.3.0) (2019-06-07)
 
 **Added**
 
--- a/src/coordinator/roles.js
+++ b/src/coordinator/roles.js
@@ -11,6 +11,24 @@ import { toSnakeCase, toCamelCase } from '../utils/objects';
  * @property {ContxtStack[]} stacks
  * @property {string} updatedAt ISO 8601 Extended Format date/time string
  */
+/**
+ * @typedef {Object} ContxtRoleApplication
+ * @property {number} applicationId
+ * @property {string} createdAt ISO 8601 Extended Format date/time string
+ * @property {number} id
+ * @property {string} roleId
+ * @property {string} updatedAt ISO 8601 Extended Format date/time string
+ */
+
+/**
+ * @typedef {Object} ContxtRoleStack
+ * @property {string} accessType Access Type of the user for this stack with options "reader", "collaborator", "owner"
+ * @property {string} createdAt ISO 8601 Extended Format date/time string
+ * @property {number} id
+ * @property {string} userId
+ * @property {number} stackId
+ * @property {string} updatedAt ISO 8601 Extended Format date/time string
+ */
 
 /**
  * @typedef {Object} ContxtStack
@@ -54,15 +72,15 @@ class Roles {
    *
    * @param {string} organizationId The ID of the organization
    * @param {string} roleId The UUID formatted ID of the role
-   * @param {string} applicationId The UUID formatted ID of the application
+   * @param {number} applicationId The ID of the application
    *
    * @returns {Promise}
-   * @fulfill {undefined}
+   * @fulfill {ContxtRoleApplication}
    * @reject {Error}
    *
    * @example
    * contxtSdk.roles
-   *   .addApplication('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58')
+   *   .addApplication('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '58')
    *   .then((roleApplication) => console.log(roleApplication))
    *   .catch((err) => console.log(err));
    */
@@ -98,16 +116,16 @@ class Roles {
    *
    * @param {string} organizationId The ID of the organization
    * @param {string} roleId The UUID formatted ID of the role
-   * @param {string} stackId The UUID formatted ID of the stack
+   * @param {number} stackId The ID of the stack
    * @param {'reader' | 'collaborator' | 'owner'} accessType The level of access for the role
    *
    * @returns {Promise}
-   * @fulfill {undefined}
+   * @fulfill {ContxtRoleStack}
    * @reject {Error}
    *
    * @example
    * contxtSdk.roles
-   *   .addStack('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58', 'collaborator')
+   *   .addStack('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '58', 'collaborator')
    *   .then((roleStack) => console.log(roleStack))
    *   .catch((err) => console.log(err));
    */
@@ -270,7 +288,7 @@ class Roles {
    *
    * @param {string} organizationId The ID of the organization
    * @param {string} roleId The UUID formatted ID of the role
-   * @param {string} applicationId The UUID formatted ID of the application
+   * @param {number} applicationId The ID of the application
    *
    * @returns {Promise}
    * @fulfill {undefined}
@@ -278,8 +296,7 @@ class Roles {
    *
    * @example
    * contxtSdk.roles
-   *   .removeApplication('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58')
-   *   .then((roleApplication) => console.log(roleApplication))
+   *   .removeApplication('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '58')
    *   .catch((err) => console.log(err));
    */
   removeApplication(organizationId, roleId, applicationId) {
@@ -314,7 +331,7 @@ class Roles {
    *
    * @param {string} organizationId The ID of the organization
    * @param {string} roleId The UUID formatted ID of the role
-   * @param {string} stackId The UUID formatted ID of the stack
+   * @param {number} stackId The ID of the stack
    *
    * @returns {Promise}
    * @fulfill {undefined}
@@ -322,8 +339,7 @@ class Roles {
    *
    * @example
    * contxtSdk.roles
-   *   .removeStack('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58')
-   *   .then((roleStack) => console.log(roleStack))
+   *   .removeStack('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '58')
    *   .catch((err) => console.log(err));
    */
   removeStack(organizationId, roleId, stackId) {
--- a/src/coordinator/roles.js
+++ b/src/coordinator/roles.js
@@ -99,7 +99,7 @@ class Roles {
    * @param {string} organizationId The ID of the organization
    * @param {string} roleId The UUID formatted ID of the role
    * @param {string} stackId The UUID formatted ID of the stack
-   * @param {'reader' | 'collaborator' | 'owner'} [accessType='reader'] The level of access for the role
+   * @param {'reader' | 'collaborator' | 'owner'} accessType The level of access for the role
    *
    * @returns {Promise}
    * @fulfill {undefined}
@@ -131,7 +131,11 @@ class Roles {
     }
 
     if (['reader', 'collaborator', 'owner'].indexOf(accessType) === -1) {
-      accessType = 'reader';
+      return Promise.reject(
+        new Error(
+          'An accessType of "reader", "collaborator", or "owner" is required for adding a stack to a role.'
+        )
+      );
     }
 
     return this._request
--- a/src/coordinator/roles.js
+++ b/src/coordinator/roles.js
@@ -62,8 +62,8 @@ class Roles {
    *
    * @example
    * contxtSdk.roles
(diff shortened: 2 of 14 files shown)


src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
     this._sdk = sdk;
   }
 
+  /**
+   * Add an application to a role
+   *
+   * API Endpoint: '/applications/:applications_id/roles/:roleId'
+   * Method: POST
+   *
+   * @param {string} organizationId The ID of the organization
```
smstromb: Are you just being proactive with asking for this organizationId without actually using it yet?

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
+
+    if (!stackId) {
+      return Promise.reject(
+        new Error('A stackId is required for adding a stack to a role.')
+      );
+    }
+
+    if (['reader', 'collaborator', 'owner'].indexOf(accessType) === -1) {
+      accessType = 'reader';
+    }
```
smstromb: This'll do, but what are your thoughts on just rejecting with an error if no access type is specified?

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {string} applicationId The UUID formatted ID of the application
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .addApplication('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58', 'collaborator')
```
smstromb: You can remove the 4th argument

src/coordinator/roles.js
```diff
@@ -161,6 +257,88 @@ class Roles {
...
+   * @param {string} organizationId The ID of the organization
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {string} applicationId The UUID formatted ID of the application
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles.delete('4f0e51c6-728b-4892-9863-6d002e61204d');
```
smstromb: Needs updated to be `roles.removeApplication(....)` with the 3 args

src/coordinator/roles.js
```diff
@@ -161,6 +257,88 @@ class Roles {
...
+   * @param {string} stackId The UUID formatted ID of the stack
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles.delete('4f0e51c6-728b-4892-9863-6d002e61204d');
+   */
+  removeStack(organizationId, roleId, stackId) {
```
smstromb: Needs updated to be roles.removeStack(....) with the 3 args

src/coordinator/roles.spec.js
```diff
@@ -29,6 +29,200 @@ describe('Coordinator/Roles', function() {
...
+        const roles = new Roles(baseSdk, baseRequest, expectedHost);
+        const promise = roles.addStack(organization.id, role.id, null, roleStack.accessType);
+
+        return expect(promise).to.be.rejectedWith(
+          'A stackId is required for adding a stack to a role.'
+        );
+      });
+    });
+
+  });
```
smstromb: You should add test contexts for `when the organization ID is not provided` and another one for `when the access type is not provided` to ensure it defaults.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## [v1.2.3](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.2.3) (2019-06-05)
```
smstromb: I'm not sure if anyone else has a stronger opinion on this but I think this qualifies as a minor version bump due to just adding brand new functionality vs bug fixes.

src/coordinator/roles.js
melting: yes

src/coordinator/roles.js
melting: thx

CHANGELOG.md
melting: This is what lee told me last time...

src/coordinator/roles.js
melting: done

src/coordinator/roles.js
melting: done

CHANGELOG.md
leeallen337: I don't think we've settled on any clear internal guidelines to what should be a patch bump vs a minor bump. In the past we've always just done a patch but we should probably set some clear guidelines sometime soon

src/coordinator/roles.spec.js
```diff
@@ -29,6 +29,200 @@ describe('Coordinator/Roles', function() {
...
+
+        it('formats the returned role stack', function() {
+          return promise.then(() => {
+            expect(toCamelCase).to.be.calledWith(roleStackFromServer);
+          });
+        });
+
+        it('returns a fulfilled promise', function() {
+          return expect(promise).to.be.fulfilled;
+        });
```
leeallen337: Probably needs a test about it actually returning the role stack?

CHANGELOG.md
greg5green: It should be a minor bump:
- MAJOR version when you make incompatible API changes,
- MINOR version when you add functionality in a backwards-compatible manner, and
- PATCH version when you make backwards-compatible bug fixes.

leeallen337: We should probably think about how we're going to handle patching a minor version that we've already passed.

For example if `master` is on `1.4.1` and someone sees a bug in `1.2.2` what's the git/work flow to patch `1.2.2`?

greg5green: Probably just patch it once per major version - a minor release shouldn't cause any breaking changes so I don't see it as a problem to update the most recent version

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
+   *
+   * API Endpoint: '/applications/:applications_id/roles/:roleId'
+   * Method: POST
+   *
+   * @param {string} organizationId The ID of the organization
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {string} applicationId The UUID formatted ID of the application
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
```
greg5green: Is this really undefined? If it is, you should leave off the `console.log(roleApplication)` in the example, and skip the `.then`

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .addStack('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58', 'collaborator')
+   *   .then((roleStack) => console.log(roleStack))
+   *   .catch((err) => console.log(err));
+   */
+  addStack(organizationId, roleId, stackId, accessType) {
```
greg5green: What do you think about passing an object in here (and maybe the other methods too)? In the code that is invoked (and/or the example), having 3 UUIDs and a string isn't the easiest to read

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
+   * API Endpoint: '/applications/:applications_id/stacks/:stackId'
+   * Method: POST
+   *
+   * @param {string} organizationId The ID of the organization
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {string} stackId The UUID formatted ID of the stack
+   * @param {'reader' | 'collaborator' | 'owner'} [accessType='reader'] The level of access for the role
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
```
greg5green: Same question here about undefined vs. example/formatting to camelcase in the .then

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
+    }
+
+    if (['reader', 'collaborator', 'owner'].indexOf(accessType) === -1) {
+      accessType = 'reader';
+    }
+
+    return this._request
+      .post(
+        `${this._baseUrl}/stacks/${stackId}/roles/${roleId}`,
+        {access_type: accessType}
```
greg5green: 1. We try to use `toSnakeCase` for this type of thing
2. I'm really surprised prettier/pretty-quick didn't format this differently -- can you make sure it runs?

src/coordinator/roles.js
```diff
@@ -161,6 +257,94 @@ class Roles {
...
+   * @param {string} applicationId The UUID formatted ID of the application
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .removeApplication('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58')
+   *   .then((roleApplication) => console.log(roleApplication))
```
greg5green: It's usually a 204 on delete, so I'd remove this from the docs

src/coordinator/roles.js
```diff
@@ -161,6 +257,94 @@ class Roles {
...
+   * @param {string} stackId The UUID formatted ID of the stack
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .removeStack('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58')
+   *   .then((roleStack) => console.log(roleStack))
```
greg5green: Same comment as the other remove method about a 204

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .addApplication('4f0e51c6-728b-4892-9863-6d002e61204d', '36b8421a-cc4a-4204-b839-1397374fb16b', '007ca9ee-ece7-4931-9d11-9b4fd97d4d58')
+   *   .then((roleApplication) => console.log(roleApplication))
+   *   .catch((err) => console.log(err));
+   */
+  addApplication(organizationId, roleId, applicationId) {
+    if (!organizationId) {
```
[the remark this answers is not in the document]
melting: Currently it is not.  Since this is a public I don't want to introduce a breaking change later.

src/coordinator/roles.js
melting: done

src/coordinator/roles.spec.js
melting: done

CHANGELOG.md
melting: > We should probably think about how we're going to handle patching a minor version that we've already passed.
> 
> For example if `master` is on `1.4.1` and someone sees a bug in `1.2.2` what's the git/work flow to patch `1.2.2`?

Usually at that point you can branch off of the tagged change.

src/coordinator/roles.js
melting: done

src/coordinator/roles.spec.js
melting: done

src/coordinator/roles.js
melting: done

src/coordinator/roles.js
```diff
@@ -161,6 +257,94 @@ class Roles {
...
+
+  /**
+   * Remove an stack from a role
+   *
+   * API Endpoint: '/stacks/:stacks_id/roles/:roleId'
+   * Method: DELETE
+   *
+   * @param {string} organizationId The ID of the organization
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {string} stackId The UUID formatted ID of the stack
```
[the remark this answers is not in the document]
melting: thanks! Done

src/coordinator/roles.js
```diff
@@ -161,6 +257,94 @@ class Roles {
...
+
+  /**
+   * Remove an application from a role
+   *
+   * API Endpoint: '/applications/:applications_id/roles/:roleId'
+   * Method: DELETE
+   *
+   * @param {string} organizationId The ID of the organization
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {string} applicationId The UUID formatted ID of the application
```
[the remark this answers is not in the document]
melting: thanks...done

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
+
+  /**
+   * Add a stack to a role
+   *
+   * API Endpoint: '/applications/:applications_id/stacks/:stackId'
+   * Method: POST
+   *
+   * @param {string} organizationId The ID of the organization
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {string} stackId The UUID formatted ID of the stack
```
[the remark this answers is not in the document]
melting: thanks....done

src/coordinator/roles.js
```diff
@@ -46,6 +46,102 @@ class Roles {
...
 
+  /**
+   * Add an application to a role
+   *
+   * API Endpoint: '/applications/:applications_id/roles/:roleId'
+   * Method: POST
+   *
+   * @param {string} organizationId The ID of the organization
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {string} applicationId The UUID formatted ID of the application
```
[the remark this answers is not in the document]
melting: thanks...done

src/coordinator/roles.js
melting: thx...done

src/coordinator/roles.js
melting: thx...done

src/coordinator/roles.js
melting: thx...done

src/coordinator/roles.js
melting: I pulled out the organizationId.  This is now 100% same api roles -> users.  I think this makes it pretty clean.

src/coordinator/roles.js
melting: 1. I think this is more explicit when there is no object to run on, and it matches users.
2. Prettier is running...

greg5green: w/r/t to 2, why would it reformat differently if i was to run it?

src/coordinator/roles.spec.js
```diff
@@ -29,6 +29,200 @@ describe('Coordinator/Roles', function() {
...
+
+      it('formats the returned role application', function() {
+        return promise.then(() => {
+          expect(toCamelCase).to.be.calledWith(roleApplicationFromServer);
+        });
+      });
+
+      it('returns a fulfilled promise', function() {
+        return expect(promise).to.be.fulfilled;
+      });
```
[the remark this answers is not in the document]
melting: @leeallen337 perhaps I should just not return the mapping data.  Is it useful?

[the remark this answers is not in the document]
leeallen337: Generally speaking, all our `POST` endpoint return the item created, even if you may just be posting to some type of join table or the data isn't necessarily useful, so I don't know if it would be a good idea to deviate from that pattern. It might be a little inconsistent to have some `POST`s return data while others don't

You could wrap it into one test (I think you can chain it like below) and do something like
```
it('returns a fulfilled promise with the data', function() {
  return expect(promise).to.be.fulfilled
    .and.to.eventually.deep.equal(expectedRoleApplication);
});
```
Maybe that works? What do you think?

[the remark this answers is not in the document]
melting: dones

src/coordinator/roles.spec.js
melting: done

melting:
> I have a few questions and requests.
> 
> Also maybe i'm being very picky but when passing in multiple arguments like
> 
> ```
> addApplication(organizationId, roleId, applicationId)
> ```
> 
> could we make the order match the order they show up in the request url so for this one since the url is
> 
> ```
> /applications/${applicationId}/roles/${roleId}
> ```
> 
> the function definition would be
> 
> ```
> addApplication(organizationId, applicationId, roleId)*
> ```
> 
> That would give us an easy and clear standard that we could use for these use cases.
> 
> Thoughts?
> 
> * I'm assuming that organizationId would come first, since it isn't in the url yet

I purposed didn't go that route...
1) I think that the role has a set of permissions, so I added this method from the perspective of the role.  If we want to be fully mimic the api, this method should go to the stack and application modules.
2) I don't think SDK's should always map 1 to 1 to their api's.  If this is the case then users could just call the api directly.
3) I wanted to mimic the permissions for the users

If you feel strongly, I will defer to you on this.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## [v1.3.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.3.0) (2019-06-06)
```
leeallen337: May need to update this to `2019-06-07`

src/coordinator/roles.js
```diff
@@ -46,6 +64,93 @@ class Roles {
...
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {number} applicationId The ID of the application
+   *
+   * @returns {Promise}
+   * @fulfill {ContxtRoleApplication}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .addApplication('36b8421a-cc4a-4204-b839-1397374fb16b', '58')
```
leeallen337: I think `58` here should be a number rather than a string since you specify it as a number on line 74.

Realistically, it shouldn't really matter since it's getting turned into a string anyways for the request url but just to be consistent with what's listed as a `@param`. What are your thoughts?

src/coordinator/roles.js
```diff
@@ -46,6 +64,93 @@ class Roles {
...
+   * @param {number} stackId The ID of the stack
+   * @param {'reader' | 'collaborator' | 'owner'} accessType The level of access for the role
+   *
+   * @returns {Promise}
+   * @fulfill {ContxtRoleStack}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .addStack('36b8421a-cc4a-4204-b839-1397374fb16b', '58', 'collaborator')
```
leeallen337: Same comment about `58` being a number

src/coordinator/roles.js
```diff
@@ -161,6 +266,82 @@ class Roles {
...
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {number} applicationId The ID of the application
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .removeApplication('36b8421a-cc4a-4204-b839-1397374fb16b', '58')
```
leeallen337: Same request about string vs number

src/coordinator/roles.js
```diff
@@ -161,6 +266,82 @@ class Roles {
...
+   * @param {string} roleId The UUID formatted ID of the role
+   * @param {number} stackId The ID of the stack
+   *
+   * @returns {Promise}
+   * @fulfill {undefined}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.roles
+   *   .removeStack('36b8421a-cc4a-4204-b839-1397374fb16b', '58')
```
leeallen337: Same request about string vs number

CHANGELOG.md
melting: done

src/coordinator/roles.js
melting: done

src/coordinator/roles.js
melting: done

src/coordinator/roles.js
melting: done

PR #117, opened 2019-06-05
Update internal config to include webSocket key

zachary-kuhn:
## Why?

- The `webSocket` key was not included when passed to the module

## What changed?

- Potentially add `webSocket` key in the config builder

### DIFF
--- a/src/config/index.js
+++ b/src/config/index.js
@@ -8,6 +8,7 @@ import defaultConfigs from './defaults';
  * @param {string} config.clientId Client Id provided by Auth0 for the environment you are
  *   trying to communicate with
  * @param {string} config.host Hostname for the API that corresponds with the clientId provided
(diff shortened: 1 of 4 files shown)


src/config/index.js
```diff
@@ -114,6 +116,7 @@ class Config {
    * clientId and host, or an environment that matches a default audience/environment.
    *
    * @param {CustomAudience} config A custom audience configuration to parse
+   * @param {Object.<string, Audience>} audiences An object with keys for environment names and values of Audience information
```
greg5green: doh, i can't believe i left that off

zachary-kuhn: Don't beat yourself up. If it weren't for me using the ultra cool IntelliJ IDEA Ultimate IDE, I'd have never seen the warning that the property was undocumented.

PR #118, opened 2019-06-13
Feat/subscribe to the message bus

zachary-kuhn:
## Why?

- The end user needs to be able to subscribe to Message Bus channels

## What changed?

- Add subscribe method for listening to messages from a channel
- Add private acknowledge method for acknowledging messages from the channel
- Refactor single use JSON RPC listeners
- Refactor is connected check
- Test new methods
- Document new methods

### DIFF
--- a/src/bus/webSocketConnection.js
+++ b/src/bus/webSocketConnection.js
@@ -175,13 +175,25 @@ class WebSocketConnection {
   }
 
   /**
-   * Subscribes to a specific channel on the message bus and handles messages as they are received
+   * Subscribes to a specific channel on the message bus and handles messages as they are received. When the handler is
+   * called, the message is automatically acknowledged after the message completes except whenever an Error is thrown.
+   * The user can also programmatically control when the message is acknowledged by calling `ack` at any time.
    *
    * @param {string} serviceClientId Client ID of the message bus service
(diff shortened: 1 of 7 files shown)


src/bus/webSocketConnection.js
```diff
@@ -173,32 +146,196 @@ class WebSocketConnection {
...
+      return Promise.reject(
+        new Error('A service client id is required for subscribing')
+      );
+    }
+
+    if (!channel) {
+      return Promise.reject(new Error('A channel is required for subscribing'));
+    }
+
+    if (typeof group === 'function') {
```
erichardson30: Just from an organizational standpoint, what do you think about moving this down below all the other `error` checking? Just so you can read through all the error cases quickly and then when all those pass, you get to the actionable stuff

zachary-kuhn: You're right! Moving it below would affect how `handler` and `errorHandler` are defined, but I can move it to the top. The code is to allow `group` to be a completely optional parameter, but not at the end of the function because an API like `subscribe(serviceId, channel, (message) => {}, (error) => {}, group)` looks a little odd to me. Great suggestion, should help the readability a bit.

erichardson30: Ah, yes. That's right 👍

PR #119, opened 2019-06-13
Bump/update dependencies with vulnerabilities

zachary-kuhn:
## Why?

- Our versions of Axios and Mocha include dependencies with known vulnerabilities

## What changed?

- Bump version of Axios
- Bump version of Mocha


zachary-kuhn:
Note, tests have been run which hopefully ensures the Mocha library is compatible. I've browsed the Axios changelog and don't see any changes between 0.17 and 0.19 which would likely affect our use. Because of this, this change isn't zero risk but will hopefully be low risk.

PR #120, opened 2019-06-14
Bump Sinon dependencies

zachary-kuhn:
## Why?

- The version of Sinon we use includes a dependency with a known vulnerability

## What changed?

- Update sinon dependency
- Update sinon-chai dependency
- Refactor code to not use deprecated method
- Make changes to linting and the test harness to make this work

### DIFF
--- a/src/utils/pagination/formatPaginatedDataFromServer.spec.js
+++ b/src/utils/pagination/formatPaginatedDataFromServer.spec.js
@@ -4,7 +4,7 @@ import formatPaginatedDataFromServer from './formatPaginatedDataFromServer';
 
 describe('utils/pagination/formatPaginatedDataFromServer', function() {
   beforeEach(function() {
-    this.sandbox = sandbox.create();
+    this.sandbox = sinon.createSandbox();
   });
 
   afterEach(function() {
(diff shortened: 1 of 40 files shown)


zachary-kuhn:
Worth considering, the `sinon` documentation includes this paragraph:

> Default sandbox
> 
> Since sinon@5.0.0, the sinon object is a default sandbox. Unless you have a very advanced setup or need a special configuration, you probably want to just use that one.

Which to me implies we should maybe just use the `sinon` object, but @greg5green knows way more about the why behind our setup.

src/utils/pagination/formatPaginatedDataFromServer.spec.js
```diff
@@ -3,12 +3,10 @@ import * as objectUtils from '../objects';
 import formatPaginatedDataFromServer from './formatPaginatedDataFromServer';
 
 describe('utils/pagination/formatPaginatedDataFromServer', function() {
-  beforeEach(function() {
-    this.sandbox = sinon.createSandbox();
-  });
+  beforeEach(function() {});
```
greg5green: There's a few empty blocks like this -- can you remove them?

zachary-kuhn: lol that is pretty damn good.

PR #122, opened 2019-06-27
Dynamic Module Mounting

greg5green:
## Why?
When building web applications, there are some circumstances where it might be necessary to add an external module after the application is initially started up (i.e. packages or scripts are loaded after the initial runtime via things like dynamic imports or RequireJS).

## What changed?
- Added the ability to clear an existing API access token from the in-memory token stores for the `Auth0WebAuth` and `MachineAuth` session types
- Added the ability to dynamically add and remove an audience from the SDK's config object after the SDK's initial instantiation
- Added the ability to mount and unmount modules dynamically after the SDK's initial instantiation
  - Uses a different API than the way external modules are added at instantiation, but module objects look the same (providing the module, a client ID, and a host)

### DIFF
--- a/support/addons.js
+++ b/support/addons.js
@@ -2,7 +2,9 @@
 
 const chai = require('chai');
 const chaiAsPromised = require('chai-as-promised');
+const chaiUuid = require('chai-uuid');
 const sinonChai = require('sinon-chai');
 
 chai.use(chaiAsPromised);
+chai.use(chaiUuid);
 chai.use(sinonChai);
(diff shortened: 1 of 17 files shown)


jgkingston:
How would one go about testing this?

support/addons.js
```diff
@@ -2,7 +2,9 @@
 
 const chai = require('chai');
 const chaiAsPromised = require('chai-as-promised');
+const chaiUuid = require('chai-uuid');
```
erichardson30: is this being used anywhere?

greg5green: No, it is not anymore. Sloppy sloppy.

I'll remove.

greg5green:
@jgkingston I wired this up in nSight to confirm it works as expected 👍 . Merging since it did!

PR #123, opened 2019-06-28
Adds the consent module to coordinator

smstromb:
## Why?
A contxt application needs to be able to verify if a user's consent is required and accept the consent form if necessary.  

## What changed?

- Added `Consent` module to `Coordinator` with typedefs.
- Added `Consent#accept`
   - Accepts a user's consent to an application
```js
contxtSdk.coordinator.consent
   .accept('36b8421a-cc4a-4204-b839-1397374fb16b')
   .then((userApproval) => console.log(userApproval))
   .catch((err) => console.log(err));
```

- Added `Consent#verify`
  - Gets the current application version's consent forms. The current access_token will be used to derive which application is being consented to.
```js
contxtSdk.coordinator.consent
   .get()
   .then((applicationConsent) => console.log(applicationConsent))
   .catch((err) => console.log(err));
```

### DIFF
--- a/src/coordinator/consent.js
+++ b/src/coordinator/consent.js
@@ -64,7 +64,7 @@ class Consent {
   }
 
   /**
-   * Accepts an application's consent for a given audience name
+   * Accepts a user's consent to an application for a given audience name
    *
    *
    * API Endpoint: '/consents/:consentId/accept'
@@ -78,7 +78,7 @@ class Consent {
    * @reject {Error}
    *
    * @example
-   * contxtSdk.consent
+   * contxtSdk.coordinator.consent
    *   .accept('coordinator', '36b8421a-cc4a-4204-b839-1397374fb16b')
    *   .then((userApproval) => console.log(userApproval))
    *   .catch((err) => console.log(err));
@@ -127,7 +127,7 @@ class Consent {
    * @reject {Error}
    *
    * @example
-   * contxtSdk.consent
+   * contxtSdk.coordinator.consent
    *   .verify('coordinator')
    *   .then((applicationConsent) => console.log(applicationConsent))
    *   .catch((err) => console.log(err));
--- a/src/coordinator/consent.js
+++ b/src/coordinator/consent.js
@@ -120,11 +120,11 @@ class Consent {
    *
    * @example
    * contxtSdk.coordinator.consent
-   *   .get()
+   *   .getForCurrentApplication()
    *   .then((applicationConsent) => console.log(applicationConsent))
    *   .catch((err) => console.log(err));
    */
-  get() {
+  getForCurrentApplication() {
     return this._sdk.auth.getCurrentAccessToken().then((accessToken) => {
       if (!accessToken) {
         return Promise.reject(new Error('A valid JWT token is required'));
--- a/src/coordinator/consent.js
+++ b/src/coordinator/consent.js
@@ -91,17 +91,17 @@ class Consent {
       );
     }
 
-    const { accessToken } = this._sdk.auth._getStoredSession();
+    return this._sdk.auth.getCurrentAccessToken().then((accessToken) => {
+      if (!accessToken) {
+        return Promise.reject(new Error('A valid JWT token is required'));
+      }
 
-    if (!accessToken) {
-      return Promise.reject(new Error('A valid JWT token is required'));
-    }
-
-    return this._request
-      .post(`${this._baseUrl}/consents/${consentId}/accept`, {
-        access_token: accessToken
-      })
-      .then((userApproval) => toCamelCase(userApproval));
+      return this._request
+        .post(`${this._baseUrl}/consents/${consentId}/accept`, {
+          access_token: accessToken
+        })
+        .then((userApproval) => toCamelCase(userApproval));
+    });
   }
 
   /**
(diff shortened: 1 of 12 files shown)


src/coordinator/consent.js
```diff
@@ -0,0 +1,137 @@
...
+   *
+   * @param {string} consentId The ID of the consent form the user is accepting
+   *
+   * @returns {Promise}
+   * @fulfill {ContxtUserConsentApproval}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.coordinator.consent
+   *   .accept('coordinator', '36b8421a-cc4a-4204-b839-1397374fb16b')
```
greg5green: This differs from the method's accepted parameters

src/coordinator/consent.js
```diff
@@ -0,0 +1,137 @@
...
+   * API Endpoint: '/applications/consent'
+   * Method: POST
+   *
+   * @returns {Promise}
+   * @fulfill {ContxtApplicationConsent}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.coordinator.consent
+   *   .verify('coordinator')
```
greg5green: This differs from the method's accepted parameters too

src/coordinator/consent.js
```diff
@@ -0,0 +1,137 @@
...
+  verify() {
+    const { accessToken } = this._sdk.auth._getStoredSession();
+
+    if (!accessToken) {
+      return Promise.reject(new Error('A valid JWT token is required'));
+    }
+
+    return this._request
+      .post(`${this._baseUrl}/applications/consent`, {
+        access_token: accessToken
```
greg5green: This will only ever get the consent tree around whatever app this method is used from -- is this the intent? Can we get some further documentation around this?

src/coordinator/consent.js
```diff
@@ -0,0 +1,137 @@
...
+   * @fulfill {ContxtApplicationConsent}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.coordinator.consent
+   *   .verify('coordinator')
+   *   .then((applicationConsent) => console.log(applicationConsent))
+   *   .catch((err) => console.log(err));
+   */
+  verify() {
```
greg5green: It seems like this returns a value instead of a boolean -- there's a little cognitive dissonance in my head over this

greg5green: The name of this isn't a deal breaker for me, btw. Just raising the issue

smstromb:
> 1. Instead of using `_getSessionInfo()` which pulls from local storage, could you create a `getCurrentAccessToken` method which pulls that from `_sessionInfo`? We had one until I was refactoring and pulled it out in #103 😬

This was originally the case in my first commit: https://github.com/ndustrialio/contxt-sdk-js/pull/123/commits/a59b75afd8dc341e4564443365c1a771522c1a42

However, unfortunately the accessToken stored in in localStorage and returned with `_getSessionInfo()` is NOT the same accessToken stored in `_sessionInfo` and returned by `getCurrentAccessToken`. 

> 2. I'd like to see documentation about how this only works from the auth0WebAuth session type (and I think the `passwordGrant` type too). Or maybe the documentation should be that it doesn't work in the `machineAuth` type

Machine's are not legal entities as far as I'm aware, but I'll document :+1:

src/coordinator/consent.js
smstromb: > is this the intent?

I don't know what the intent is or was for these endpoints, I just learned about these a few days ago when I was told to move it as-is from contxt 1.0 to contxt 2.0. The endpoints remain as they were, the only change is adding them to the sdk. 

I could attempt to document the reasoning behind extracting the active application ID out of the JWT token instead of explicitly asking based on applicaiton ID, but to be honest, I'd most likely be wrong.

src/coordinator/consent.js
smstromb: makes sense, I'll change to `consent#get`

greg5green:
>This was originally the case in my first commit: a59b75a
>
>However, unfortunately the accessToken stored in localStorage and returned with _getSessionInfo() is NOT the same accessToken stored in _sessionInfo and returned by getCurrentAccessToken.

Yes, you were using `getCurrent**Api**Token`, which returns a promise that resolves with an API token for a particular API.

To add clarity, I'm saying we should add a `getCurrentAccessToken` that provides a stable non-private API for getting the access token (which, on an implementation level, would be better to get from `_sessionInfo` as it wouldn't be dependent on the way we are choosing to store the token between page loads/sessions/etc).

smstromb:
> > This was originally the case in my first commit: [a59b75a](https://github.com/ndustrialio/contxt-sdk-js/commit/a59b75afd8dc341e4564443365c1a771522c1a42)
> > However, unfortunately the accessToken stored in localStorage and returned with _getSessionInfo() is NOT the same accessToken stored in _sessionInfo and returned by getCurrentAccessToken.
> 
> Yes, you were using `getCurrent**Api**Token`, which returns a promise that resolves with an API token for a particular API.
> 
> To add clarity, I'm saying we should add a `getCurrentAccessToken` that provides a stable non-private API for getting the access token (which, on an implementation level, would be better to get from `_sessionInfo` as it wouldn't be dependent on the way we are choosing to store the token between page loads/sessions/etc).

gotcha gotcha, I've commited this suggestion

smstromb:
> This looks good. Only small concern I have is with the `get` method name now -- should it be more specific about what consents it is getting? In case the API changes it's behavior down the line? Probably not without a big change, so :shipit:

@greg5green Are you good with `consent.getByAccessToken()`?

greg5green:
That kind of implies you can provide the access token. What about `getForCurrentApplication` or something?

PR #124, opened 2019-07-12
Fix how auth0-js is imported in Session Types

greg5green:
## Why?
`auth0-js` recently changed code that is run when requiring/importing the module to require `window` to be available before any methods are invoked. This was causing issues in Node environments, even if a session type that used `auth0-js` was not used.

## What changed?
- Added methods that require/import the individual Session Types only when attempting to use them
- Updated the version of `lodash` used by dependencies (I think only `devDependencies`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [v1.7.1](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.7.1) (2019-07-12)
+
+**Changed**
+
+- Updated how session types are required/imported into the SDK
+  - Solves a problem where `auth0-js` required `window` in Node environments
+
 ## [v1.7.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.7.0) (2019-07-03)
 
 **Added**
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,11 +1,18 @@
...
+
 ## [v1.7.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.7.0) (2019-07-03)
 
 **Added**
 
 - Added `Coordinator.consent#getForCurrentApplication` for getting the current application version's consent form. The current access_token will be used to derive which application is being consented to.
 - Added `Coordinator.consent#accept` for accepting user consent to an applications
 
-## [v1.6.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.5.0) (2019-06-25)
+## [v1.6.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v1.6.0) (2019-06-25)
```
leeallen337: 💯

src/sessionTypes/index.spec.js
```diff
@@ -0,0 +1,102 @@
+import auth0 from 'auth0-js';
+import Auth0WebAuth from './auth0WebAuth';
+import PasswordGrantAuth from './passwordGrantAuth';
+import MachineAuth from './machineAuth';
+import * as sessionTypes from './index';
+
+describe('sessionTypes', function() {
+  afterEach(function() {
+    sinon.restore();
+  });
```
leeallen337: Is there no need for a `beforeEach` here to set up `sinon`?

greg5green: Zach pointed out a change in `sinon` about a month ago, where newer versions have a default sandbox set up already! https://github.com/ndustrialio/contxt-sdk-js/pull/120#issuecomment-502125529

leeallen337: Cool cool cool

grettaface: This is cool!

PR #126, opened 2019-07-22
Adds auth0 full sync endpoint

smstromb:
## Why?

https://developers.ndustrial.io (contxt 1.0) calls `GET /users/:userId/sync` upon login to force a user's role and application access to sync to auth0. This is done because there's a number of contxt-coordinator endpoints that are responsible for updating user access but defer updating that access in Auth0, instead, relying on the `/sync` endpoint as a catch-all for auth0-contxt sync issues. 

https://contxt.ndustrial.io (contxt 2.0) currently does NOT call `GET /users/:userId/sync`. 

It needs to.

## What changed? (v2.1.0)

- Added `Coordinator.users#syncWithAuth0(userId)`

### DIFF
--- a/src/coordinator/users.js
+++ b/src/coordinator/users.js
@@ -512,7 +512,7 @@ class Users {
   }
 
   /**
-   * Syncs the user's roles, servies, and application access with auth0
+   * Syncs the user's roles and application access with auth0
    *
    * API Endpoint: '/users/:userId/sync'
    * Method: GET
@@ -525,7 +525,7 @@ class Users {
    *
    * @example
    * contxtSdk.coordinator.users
-   *   .sync('36b8421a-cc4a-4204-b839-1397374fb16b')
+   *   .syncWithAuth0('36b8421a-cc4a-4204-b839-1397374fb16b')
    *   .catch((err) => console.log(err));
    */
   syncWithAuth0(userId) {
--- a/src/coordinator/users.js
+++ b/src/coordinator/users.js
@@ -510,6 +510,33 @@ class Users {
       `${this._baseUrl}/users/${userId}/stacks/${stackId}`
     );
   }
+
+  /**
+   * Syncs the user's roles, servies, and application access with auth0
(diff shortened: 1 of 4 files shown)


src/coordinator/users.js
```diff
@@ -510,6 +510,33 @@ class Users {
...
+   *   .catch((err) => console.log(err));
+   */
+  syncWithAuth0(userId) {
+    if (!userId) {
+      return Promise.reject(
+        new Error('A user ID is required for syncing with auth0')
+      );
+    }
+
+    return this._request.get(`${this._baseUrl}/users/${userId}/sync`);
```
rob5standingby: It's so strange to me that this was built in Contxt Coordinator API as a GET request.

smstromb: ‾\\\_(ツ)_/‾

src/coordinator/users.js
```diff
@@ -510,6 +510,33 @@ class Users {
...
+   * @returns {Promise}
+   * @fulfill {undefined}
+   * @reject {Error}
+   *
+   * @example
+   * contxtSdk.coordinator.users
+   *   .syncWithAuth0('36b8421a-cc4a-4204-b839-1397374fb16b')
+   *   .catch((err) => console.log(err));
+   */
+  syncWithAuth0(userId) {
```
melting: Let's leave out Auth0 from name, some thoughts:
- sync
- syncWithAuthProvider
- syncPermissions

greg5green: This is good and I take back my approval

smstromb: But y tho?

This endpoint get's deleted the moment auth0 isn't used anymore.

melting: Mostly because it is an implementation detail.

Other reasons:

1. auth0 could change
2. code stays around longer that we want
3. if there isn't a good reason to change, we can agree with the API when the call is 1:1 to an API

PR #127, opened 2019-07-25
Users#activate no longer uses request interceptor for authorization header 

smstromb:
## Why?
The endpoint for `coordinator.users#activate` does not require authentication but because it uses the base request object passed into the sdk, it attempts to use the axios authorization interceptors anyway. This causes a 401 error thrown from the sdk for a user who isn't authenticated (i.e, people using the user activation link)

## What changed?
`coordinator.users#activate` uses axios directly to make the request and bypasses the interceptors used on the request object.

### DIFF
--- a/src/coordinator/users.spec.js
+++ b/src/coordinator/users.spec.js
@@ -58,7 +58,6 @@ describe('Coordinator/Users', function() {
       let promise;
       let request;
       let toSnakeCase;
-      let axiosStub;
 
       beforeEach(function() {
         user = fixture.build('contxtUser');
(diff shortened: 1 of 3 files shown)


rob5standingby:
Nice find, @smstromb !

src/coordinator/users.spec.js
```diff
@@ -93,7 +97,7 @@ describe('Coordinator/Users', function() {
       });
 
       it('posts the new user to the server', function() {
-        expect(request.post).to.be.calledWith(
+        expect(axiosStub).to.be.calledWith(
```
leeallen337: Just FYI, I think you could drop the `axiosStub` variable and just have

`sinon.stub(axios, 'post').resolves();`

on its own and on this line you could have

`expect(axios.post).to.be.calledWith(`

and it would work just the same

smstromb: Awesome tip, thanks!

PR #129, opened 2019-09-09
Add ability to pass options to Auth0WebAuth#logOut

greg5green:
## Why?
We need the ability pass options (specifically `federated=true`) to Auth0 to fully set up SLO (single log out).

## What changed?
- Added `options` param to `Auth0WebAuth#logOut`


greg5green:
It does not work -- but it actually tries to make the request with the
flag. Without the flag, it doesn't even try.

PR #131, opened 2019-10-01
MOB-3121: Update coordinator modules to use new API endpoints

leeallen337:
## Why?

https://lineagelogistics.atlassian.net/browse/MOB-3121

## What changed?
- added a `setOrganizationId` method to the base `coordinator` module
  - `contxtSdk.coordinator.setOrganizationId`
  - setting an organization id on the coordinator module will then re-instantiate the coordinator sub-modules to use the appropriate tenancy-base urls
  - setting the organization id to `null` will then reset all the urls to the legacy endpoints
- updated all the coordinator submodules to accept the selected organization id and use the new tenancy-base urls (stripping out `/organizations/:organizationId` where appropriate
- kept the function API's basically the same
- updated tests

#### CHANGELOG: I went with a minor version bump since there is no breaking change on the API surface. However, I could also see this being a major version bump. Let me know what you think

### DIFF
--- a/src/coordinator/applications.spec.js
+++ b/src/coordinator/applications.spec.js
@@ -1,7 +1,7 @@
 import Applications from './applications';
 import * as objectUtils from '../utils/objects';
 
-describe.only('Coordinator/Applications', function() {
+describe('Coordinator/Applications', function() {
   let baseRequest;
   let baseSdk;
   let expectedHost;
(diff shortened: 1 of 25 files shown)


src/coordinator/applications.spec.js
```diff
@@ -308,41 +424,64 @@ describe('Coordinator/Applications', function() {
...
         });
-      });
 
-      it('returns a fulfilled promise with the featured applications', function() {
-        return expect(promise).to.be.fulfilled.and.to.eventually.deep.equal(
-          expectedFeaturedApplications
-        );
+        it('gets the list of featured applications from the server and does not use the organization ID provided', function() {
+          expect(request.get).to.be.calledWith(
+            `${expectedHost}/applications/featured`
```
rob5standingby: Shouldn't this end up being `${expectedHost}/v1/{orgId}/applications/featured` for the tenancy urls? I don't see this being set.

leeallen337:
Note: It would be best to start reviewing with the `index.js` file rather than in alphabetical order. It'll make more sense if you look at that first

src/coordinator/index.js
```diff
@@ -31,6 +35,74 @@ class Coordinator {
...
+  /**
+   * Sets a selected oranization ID to be used in tenant based requests
+   *
+   * @param {string} organizationId the ID of the organization
+   *
+   * @example
+   * contxtSdk.coordinator
+   *   .setOrganizationId('36b8421a-cc4a-4204-b839-1397374fb16b');
+   */
+  setOrganizationId(organizationId) {
```
melting: When we do the next service.  Would we add those routes to coordinator?  Or would we update their service?

Perhaps we could move this up into the SDK where the clients for other services could use it.

leeallen337: Talked in-person. Leaving as-is for now

PR #135, opened 2019-12-23
Feature/asset framework list

melting:
## Why?

For the health endpoints, we need to show the latest health status, and include anything that is a descendent from integrations asset_type.

## What changed?

When fetching assets, we now have more options to expand the payload.  Including metrics, and typ descendents

### DIFF
--- a/docs/Assets.md
+++ b/docs/Assets.md
@@ -140,6 +140,8 @@ Method: GET
 | [options.assetTypeId] | <code>string</code> | UUID of the asset type to use for filtering |
 | [options.limit] | <code>Number</code> | Maximum number of records to return per query |
 | [options.offset] | <code>Number</code> | How many records from the first record to start |
+| [options.includeTypeDescendents] | <code>boolean</code> | When true will look for all asset types that are descendendents from options.assetTypeId |
+| [options.includeMetricsId] | <code>boolean</code> | If asset has an associated metric value the latest metric will be returned. |
 
 **Example**  
 ```js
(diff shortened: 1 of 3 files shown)


docs/Assets.md
```diff
@@ -140,6 +140,8 @@ Method: GET
 | [options.assetTypeId] | <code>string</code> | UUID of the asset type to use for filtering |
 | [options.limit] | <code>Number</code> | Maximum number of records to return per query |
 | [options.offset] | <code>Number</code> | How many records from the first record to start |
+| [options.includeTypeDescendents] | <code>boolean</code> | When true will look for all asset types that are descendendents from options.assetTypeId |
+| [options.includeMetricsId] | <code>boolean</code> | If asset has an associated metric value the latest metric will be returned. |
```
melting: All of these are supported together...I think there are newer options to filter by asset attributes and values.

src/assets/index.js
```diff
@@ -186,6 +186,8 @@ class Assets {
    * @param {string} [options.assetTypeId] UUID of the asset type to use for filtering
    * @param {Number} [options.limit] Maximum number of records to return per query
    * @param {Number} [options.offset] How many records from the first record to start
+   * @param {boolean} [options.includeTypeDescendents] When true will look for all asset types that are descendendents from options.assetTypeId
+   * @param {boolean} [options.includeMetricsId] If asset has an associated metric value the latest metric will be returned.
```
smstromb: This is `includeMetricId` (not plural) and it's a string

PR #144, opened 2020-02-21
Add options to bypass the seamless sso in auth0

melting:
## Why?

When an application needs to ensure users are logged out of the originating provider with "federated=true", this won't happen if they were logged back in with seamless sso.  So to prevent logging back in with seamless sso, they need to be able to force the login.

## What changed?

Added a new option `forceLogin` to the login method for auth0.  This will in turn call auth0.authorize with `prompt:login`

The intention is that when users need to have federated logout, they will also use the forceLogin.


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## [v2.11.0](http://github.com/ndustrialio/contxt-sdk-js/tree/v2.11.0) (2020-02-21)
```
leeallen337: May want to update the date

melting: done, thanks.

PR #147, opened 2020-03-12
Exposing feed types

fawcilize:
## Why?
To access the troubleshooting url on the feed type

## What changed?
Exposed a getAll endpoint for feeds

### DIFF
--- a/src/iot/feedTypes.js
+++ b/src/iot/feedTypes.js
@@ -1,3 +1,5 @@
+import { toCamelCase } from '../utils/objects';
+
 /**
  * @typedef {Object} FeedType
  * @property {String} createdAt
@@ -46,7 +48,9 @@ class FeedTypes {
    *   .catch((err) => console.log(err));
    */
   getAll() {
-    return this._request.get(`${this._baseUrl}/feeds/types`);
+    return this._request
+      .get(`${this._baseUrl}/feeds/types`)
+      .then((feedTypes) => toCamelCase(feedTypes));
   }
 }
 
--- a/src/iot/feedTypes.js
+++ b/src/iot/feedTypes.js
@@ -1,0 +1,53 @@
+/**
+ * @typedef {Object} FeedType
+ * @property {String} createdAt
+ * @property {String} downAfter
+ * @property {String} id UUID
+ * @property {String} troubleshootingUrl
+ * @property {String} type
+ * @property {String} updatedAt
+ */
+
+/**
+ * Module that provides access to feed type information
+ *
+ * @typicalname contxtSdk.iot.feedTypes
+ */
+class FeedTypes {
+  /**
+   * @param {Object} sdk An instance of the SDK so the module can communicate
+   *   with other modules
+   * @param {Object} request An instance of the request module tied to this
+   *   module's audience
+   * @param {string} baseUrl The base URL provided by the parent module
+   */
+  constructor(sdk, request, baseUrl) {
+    this._baseUrl = baseUrl;
+    this._request = request;
+    this._sdk = sdk;
+  }
+
+  /**
+   * Get a listing of all feed types
+   *
+   * API Endpoint: '/feeds/types'
+   * Method: GET
+   *
+   * @param {PaginationOptions} [paginationOptions]
+   *
+   * @returns {Promise}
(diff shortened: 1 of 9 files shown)


src/iot/index.spec.js
```diff
@@ -52,5 +53,9 @@ describe('Iot', function() {
     it('appends an instance of Outputs to the class instance', function() {
       expect(iot.outputs).to.be.an.instanceof(Outputs);
     });
+
+    it('appends the supplied FeedTypes module to the class instance', function() {
```
fawcilize: I added this test here to be consistent, but I don't see the value. We're testing a variable assignment instead of a behavior, which in my opinion is the same as testing a constant equals a value.

src/iot/feedTypes.js
```diff
@@ -0,0 +1,57 @@
+import { toCamelCase } from '../utils/objects';
+
+/**
+ * @typedef {Object} FeedType
+ * @property {String} createdAt
+ * @property {String} downAfter
```
greg5green: Can you include more information about this, like is it seconds? milliseconds? etc

src/iot/feedTypes.js
```diff
@@ -0,0 +1,57 @@
...
+ * @property {String} id UUID
+ * @property {String} troubleshootingUrl
+ * @property {String} type
```
greg5green: Can this be anything?

src/iot/feedTypes.js
```diff
@@ -0,0 +1,57 @@
...
+    this._sdk = sdk;
+  }
+
+  /**
+   * Get a listing of all feed types
+   *
+   * API Endpoint: '/feeds/types'
+   * Method: GET
+   *
+   * @param {PaginationOptions} [paginationOptions]
```
greg5green: This doesn't take PaginationOptions

src/iot/feedTypes.js
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * @returns {Promise}
+   * @fulfill {FeedTypesFromServer} Information about the feed types
```
greg5green: This returns like `[{ //feed type }]`, right? Then this should be `FeedType[]`

src/iot/feedTypes.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    expectedHost = faker.internet.url();
+
+    feedTypes = new FeedTypes(baseSdk, baseRequest, expectedHost);
+  });
+
+  afterEach(function() {
+    sinon.restore();
+  });
+
+  describe('getAll', function() {
```
greg5green: Can you refactor these to better match the other tests we have? These are very much unit tests and you're testing multiple units by relying on `toCamelCase`

Here's an example: https://github.com/ndustrialio/contxt-sdk-js/blob/master/src/facilities/costCenters.spec.js#L239

fawcilize: I still consider this a unit test, but I'm testing a behavior rather than an implementation. I think we should reserve mocking for external systems like an API, database, or separate module. That would make it an integration test.

For example, this would pass the other tests, leading to a false positive:
```
    return this._request
      .get(`${this._baseUrl}/feeds/types`)
      .then((feedTypes) => {
        const camelCased = toCamelCase(feedTypes);
        return feedTypes; // I accidentally return the wrong variable, but the test passes
      );
```

src/iot/feedTypes.js
fawcilize: Nice catch!

src/iot/feedTypes.js
fawcilize: Correctomundo

src/iot/feedTypes.js
fawcilize: I've added a comment. It's actually the name of the feed.

src/iot/feedTypes.js
fawcilize: Had to do some digging for this one lol. It turned out to be seconds. I added a comment.

src/iot/feedTypes.spec.js
greg5green: We talked through this on Slack.

PR #150, opened 2020-04-16
Bugfix/organization endpoints

leeallen337:
## Why?

The access endpoints are a little different depending on whether you are getting all organizations or just a single organization

When getting all organizations the endpoint is `/access/v1/organizations`
When getting a single organization the endpoint is `/access/v1/:organizationId`

These need to be handled a little differently

## What changed?
- Updated the organization methods on coordinator to differentiate between getting a single organization or getting all organizations
- Fixed/updated tests
- Updated the `CHANGELOG.md`
  - It hasn't been updated in a while
  - The current release is `v.2.13.0`
- Changed the pull request template so we hopefully don't fall behind on changelog updates

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,5 @@
 ## Why?
 
-
 ## What changed?
+
+- [ ] Did you update the CHANGELOG?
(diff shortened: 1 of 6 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,5 @@
 ## Why?
 
-
 ## What changed?
+
+- [ ] Did you update the CHANGELOG?
```
smstromb, line 5: guilty :grimacing:

PR #164, opened 2020-12-10
feature/update_stack_to_roject

erichardson30:
## Why?

Changes are being made to the contxt coordinator API to replace stack with projects
[JIRA](https://ndustrialio.atlassian.net/browse/CONTXT-1661)

## What changed?
- Updated stack to project
- Referencing project by slug vs stack by ID


src/coordinator/roles.spec.js
```diff
@@ -161,121 +161,129 @@ describe('Coordinator/Roles', function() {
...
-        const promise = roles.addStack(role.id, stack.id, faker.random.word());
+        const promise = roles.addProject(
+          role.id,
+          project.slug,
+          faker.random.word()
+        );
 
         return expect(promise).to.be.rejectedWith(
-          'An accessType of "reader", "collaborator", or "owner" is required for adding a stack to a role.'
+          'An accessType of "reader", "collaborator", or "owner" is required for adding a project to a role.'
```
mgagliardo91, line 286: was the work to remove `collaborator` in another ticket? I remember talking about it, but can't remember which ticket. Either way, should we include that change here?

mgagliardo91 (approved):
LGTM

src/coordinator/roles.spec.js
erichardson30: I know we talked about it as well. I'm not sure what ticket its captured in so i"ll just remove it here to be pre-emtpive

PR #170, opened 2021-01-07
Bugfix/role project environments

erichardson30:
## Why?

Roles are now related to project environments and not stacks/projects

## What changed?

For roles and users:
- addProject is now addProjectEnvironment
- removeProject is now removeProjectEnvironment
- Tests updated

Note: I updated to just a bugfix because 3.0.0 had breaking changes for new function names and those were technically broken due to the new relationships between roles & environments. This fixes that. However, if you feel a different semantic bump is needed, I will adjust


mgagliardo91 (approved):
LGTM!

PR #171, opened 2021-01-12
Bugfix/update role user env routes

erichardson30:
## Why?

- Contxt coordinator routes were updated for project environments

## What changed?

- Updated user/role add/remove project env URLs to match what they are now in contxt coordinator

### DIFF
--- a/src/coordinator/roles.js
+++ b/src/coordinator/roles.js
@@ -104,11 +104,10 @@ class Roles {
   /**
    * Add a project environment to a role
    *
-   * API Endpoint: 'projects/:projectSlug/environments/:projectEnvironmentSlug/roles/:roleId/'
+   * API Endpoint: 'roles/:roleId/project_environments/:projectEnvironmentSlug'
    * Method: POST
    *
    * @param {string} roleId The UUID formatted ID of the role
-   * @param {string} projectSlug The slug of the project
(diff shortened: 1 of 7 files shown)


src/coordinator/roles.js
```diff
@@ -411,7 +388,7 @@ class Roles {
     return this._request.delete(
       `${
         this._baseUrl
-      }/projects/${projectSlug}/environments/${projectEnvironmentSlug}/roles/${roleId}`
+      }/roles/${roleId}/project_environments/${projectEnvironmentSlug}`
```
mgagliardo91, line 391: So these two are the duplicate routes you were talking about - given that the old one was roles on an environment, should we move forward with the route for `/:environment_slug/roles/:role_id` rather than the reverse?

mgagliardo91 (commented):
LGTM - though given we have two ways to add a role to an environment, we should consider which one we should keep

src/coordinator/roles.js
erichardson30: There is now just one way to add a role to an envrionment in contxt-coordinator at this new URL. Change can be found here: https://github.com/ndustrialio/contxt-coordinator/pull/306

mgagliardo91 (approved):

PR #173, opened 2021-01-22
feat: Bump to Node v14

mgagliardo91:
## Why?

Allow repositories that use this SDK to move up in the node version.

## What changed?

- [x] Did you update the CHANGELOG?


package.json
```diff
@@ -29,8 +29,8 @@
   "license": "ISC",
   "repository": "github:ndustrialio/contxt-sdk-js",
   "engines": {
-    "npm": "^5.10.0 ^6.0.0",
-    "node": ">=6"
+    "npm": "^5.6.0",
```
greg5green, line 32: Why disallow npm 6 and 7?

greg5green (approved):

package.json
mgagliardo91: ah, I meant >, but also, why do we need to restrict on `npm` if `node` is specified?

branchvincent (commented):
looks good to me, but will defer to @greg5green

greg5green (approved):
perfect

PR #230, opened 2022-02-21
fix: set the proper hostname for staging contxt-coordinator

andreimoustache:


src/config/audiences.js
```diff
@@ -28,7 +28,7 @@ export default {
     },
     staging: {
       clientId: '8qY2xJob1JAxhmVhIDLCNnGriTM9bct8',
```
branchvincent, line 30: i **think** this needs to be updated too:
```suggestion
      clientId: 'qGzdTXcmB57zlTp86rYsivG9qEss1lbF',
```

andreimoustache: Hmmm, any idea on how I could check that?

branchvincent: this should be right, i pulled this from the api itself: https://github.com/ndustrialio/contxt-api/blob/14c3199dbb082c81b041f48b9214fc7c9ce7fbc9/charts/contxt/values-staging.yaml#L59-L60

branchvincent: to verify, we can just grab a JWT for that audience and try to make a request. the only wrinkle here is it looks like that client only exists in the staging auth0 tenant: https://github.com/ndustrialio/contxt-api/blob/14c3199dbb082c81b041f48b9214fc7c9ce7fbc9/charts/contxt/values-staging.yaml#L86-L87

branchvincent:
@andreimoustache releasing to NPM is manual right now, opened #231 to publish on merges to `main`

PR #241, opened 2022-04-11
Update sdk to use nionic-api instead of assets-api

omegafury:
## Why?

Update the js sdk to use nionic, while also removing references to the assets api in preparation for its sunsetting.

## What changed?

- Connection to tenant specific nionic instances added, as well as a nionic service to use to access the nionic-api
- Removed the assets and facilities services. Consumers of these will need to be migrated to use the nionic service.

- [ ] Did you update the CHANGELOG?


src/index.js
```diff
@@ -201,3 +199,4 @@ class ContxtSdk {
 }
 
 export default ContxtSdk;
+export { toSnakeCase, toCamelCase };
```
omegafury, line 202: This should make these functions available as first class exports, ie:
`import { toSnakeCase } from '@ndustrial/contxt-sdk'`
instead of 
`import { toSnakecase } from '@ndustrial/contxt-sdk/lib/utils/objects`

as is being done in several places in nsight and contxt-app-v2

omegafury:
it looks like the vast majority of the file changes here are from the package-lock.json

src/index.js
branchvincent: btw, you can actually do this in one line: `export { toCamelCase, toSnakeCase } from './utils/objects';`

omegafury: oh, yea, I figured it could be done a bit simpler.

branchvincent (approved):

omegafury:
It's also worth mentioning that we've been using these changes in a bunch of other repos for a few weeks already as we worked on this migration.

PR #322, opened 2023-01-31
feat: message bus onClose, onError callbacks

zebehringer:
## Why?
so that the user code can know when the connection is gone and then decide what to do

## What changed?
added optional callback functions to the connect function

- [x] Did you update the CHANGELOG?


kkikendall (approved):

zebehringer:
Lineage has a way to work around not having those callbacks, so it isn't critical. However, I do think the User-Agent header stuff is pretty useful. I just haven't spent the time to figure out how to test it

PR #338, opened 2024-03-20
chore(deps): update axios [SUP-6272]

andreimoustache:
## Why?

https://ndustrialio.atlassian.net/browse/SUP-6272


ben-fussell (approved):
@andreimoustache I agree with your concern in standup - how confident are we in these tests? I'm approving since they are passing 😅

andreimoustache:
@ben should be good, I'd say, they worked so far. If something does break, we'll just have to add some more tests around the failing thing.

PR #352, opened 2024-12-10
fix(deps): Bump dependencies to pass audit

mgagliardo91:
## What changed?

- Bump babel to new package group
- Bump eslint > v9
- Bump other deps


melting (approved):
I think this looks good.  I expect you will now need to update all the other repos to use this.  I don't know if any of these were then flagged on the npm version for people to update (or will be once we publish)

PR #402, opened 2025-07-25
CONTXT-11594 -- Address security vulnerabilities in npm-formidable

omegafury:
## Why?

- [CONTXT-11594 -- Address security vulnerabilities in npm-formidable](https://ndustrialio.atlassian.net/browse/CONTXT-11594)

## What changed?

- npm audit fix


omegafury:
I hadn't pulled the changes from main initially, so it looked like there was more to fix here. The core issue I wanted to address was that there's a low version of formidalbe being used by superagent, which is then used by auth0-js. Updating auth0-js to the highest version doesn't update those downstream dependencies like I wanted though, and I realized that the downstream dependencies of the sdk can just update their version of formidable, as there is a 2.1.5 version that fixes the vulnerability.

PR #486, opened 2026-04-14
Remove dead orgs

omegafury:
## What changed?

- removed references to vcv and waterbridge

### DIFF
--- a/src/nionic/index.js
+++ b/src/nionic/index.js
@@ -18,7 +18,6 @@ class Nionic {
       '18d8b68e-3e59-418e-9f23-47b7cd6bdd6b': 'genan',
       '02efa741-a96f-4124-a463-ae13a704b8fc': 'lineage',
       '5209751f-ea46-4b3e-a5dd-b8d03311b791': 'ndustrial',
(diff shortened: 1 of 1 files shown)


src/nionic/index.js
```diff
@@ -21,19 +21,17 @@ class Nionic {
       '2fe29680-fc3d-4888-9e9b-44be1e59c22c': 'sfnt',
```
zebehringer: what constitutes a dead org? Aren't SFNT and Trenton dead too?

zebehringer (approved):

src/nionic/index.js
omegafury: There are like ~15 dead orgs, but I've only been directed at waterbridge right now. I've also cleared up some leftover vcv referenes in 2 places that I've noticed while in the process of removing waterbridge. Unlike some of the other orgs though, I'm 100% sfnt is dead, so I'll go ahead and delete this too

omegafury: This would be more efficient if I were given a list of all the dead orgs to remove at once, because I'm having to create a ton of PRs all over the place. There would be a lot of overlap.

